PR #214, opened 2026-04-27
fix: restrict POST /api/system/restart to admins (ticket #538)

theodorecharles:
## Summary

`POST /api/system/restart` was guarded only by `requireAuth`, meaning any authenticated user — including `viewer`-role accounts — could repeatedly POST and DoS the backend by triggering `pm2 restart all` / `process.exit(0)`. This swaps the middleware to the existing `requireAdmin` so only admins can initiate a restart.

- File touched: `backend/src/routes/system.ts`
- Pattern matches other admin-only routes (`config`, `folder-management`, `metrics`, etc.) that already use `requireAdmin`.
- Behavior inside the restart handler is unchanged.

## Test plan

- [ ] Authenticated viewer hits `POST /api/system/restart` → expect `403 Access denied - admin role required`
- [ ] Authenticated manager hits `POST /api/system/restart` → expect `403`
- [ ] Authenticated admin hits `POST /api/system/restart` → expect `200 { success: true }` and restart proceeds
- [ ] Unauthenticated request → expect `401 Not authenticated`

Closes ticket #538.

### DIFF
--- a/backend/src/routes/system.ts
+++ b/backend/src/routes/system.ts
@@ -8,7 +8,7 @@ import { spawn } from 'child_process';
 import path from 'path';
 import fs from 'fs';
 import { fileURLToPath } from 'url';
-import { requireAuth } from '../auth/middleware.js';
+import { requireAdmin } from '../auth/middleware.js';
 import { error, warn, info, debug, verbose } from '../utils/logger.js';
 
 const __filename = fileURLToPath(import.meta.url);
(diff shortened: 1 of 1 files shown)


theodorecharles:

The change is a minimal, correct two-line fix: the import swaps `requireAuth` for `requireAdmin` and the route guard on `POST /restart` follows suit. `requireAdmin` returns 401 for unauthenticated callers and 403 for authenticated non-admins, satisfying both acceptance criteria. No other routes in the file are affected, the import cleanup is exact, and the pattern aligns with every other admin-only route in the codebase. No backend test suite exists to update. There is one pre-existing concern — `requireAdmin` is async but doesn't forward thrown errors to Express's `next` — but that's shared with `requireManager` and not introduced here.

PR #217, opened 2026-04-27
fix: refuse /api/setup/initialize and /upload-avatar after OOBE (ticket #533)

theodorecharles:
## Summary

Fixes ticket #533 — `POST /api/setup/initialize` (and `POST /api/setup/upload-avatar`) were mounted unauthenticated with no completed-setup guard. After the out-of-box experience an attacker who can reach the endpoint could:

- overwrite `data/config.json` (sessionSecret, authorizedEmails, OAuth credentials, allowedOrigins, allowedHosts)
- provision a brand-new admin user via `createUser({ role: 'admin' })`
- have the backend reload the rewritten config

…achieving full account takeover with no session required. A reverse proxy that exposes `/api` is sufficient.

## What changed

Adds a `refuseIfSetupComplete` middleware in `backend/src/routes/setup.ts` that returns `403` when `getConfigExists()` is true, and applies it to both `/initialize` and `/upload-avatar`. Per the ticket discussion, OOBE intentionally has no auth so the very first user can create their account — `getConfigExists()` is `false` until `/initialize` writes the file and `reloadConfig()` runs, so the OOBE call still succeeds. Every subsequent call (and any call after a normal restart with a config present) is refused. The post-OOBE avatar update path already exists at the authenticated `/api/branding/upload-avatar`.

`/status` is intentionally left open — it is read-only and the wizard depends on it.

## Test plan

- [ ] Fresh install with no `data/config.json`: complete the OOBE wizard and verify `/initialize` succeeds once.
- [ ] Replay `POST /api/setup/initialize` with the same or a different payload after OOBE — expect `403 { error: "Setup has already completed.", setupComplete: true }` and verify `data/config.json` is unchanged and no new admin user was created.
- [ ] Same replay against `POST /api/setup/upload-avatar` — expect `403` and verify `data/photos/avatar.png` is unchanged.
- [ ] Restart the backend after OOBE and confirm `/initialize` continues to return `403`.

### DIFF
--- a/backend/src/routes/setup.ts
+++ b/backend/src/routes/setup.ts
@@ -3,7 +3,7 @@
  * Handles initial setup and configuration for first-time users
  */
 
-import { Router, Request, Response } from "express";
+import { Router, Request, Response, NextFunction } from "express";
 import fs from "fs";
 import path from "path";
 import { fileURLToPath } from "url";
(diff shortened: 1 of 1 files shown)


theodorecharles:

The fix is minimal, correct, and directly satisfies the acceptance criteria. `refuseIfSetupComplete` checks the module-level `configExists` flag via `getConfigExists()` before the handler runs: on a fresh install the flag is false so OOBE succeeds and then `reloadConfig()` flips it to true; every subsequent call gets a 403. No scope creep — one file, 33 lines, both vulnerable OOBE endpoints guarded. The optional setup-token requirement is intentionally deferred, which is explicitly permitted by the ticket. No automated tests are added, but the described manual test plan is precise and the route-level behavior is straightforward to verify.

PR #218, opened 2026-04-27
fix(security): require admin role on GET /api/config (ticket #534)

theodorecharles:
## Summary

- `GET /api/config` was gated with `requireAuth`, returning the raw `data/config.json` to anyone signed in — including `viewer` and `manager` accounts. The payload contains plaintext `environment.auth.sessionSecret` (the HMAC key for `connect.sid`, so a viewer could forge admin sessions), `environment.auth.google.clientSecret`, `openai.apiKey`, `analytics.openobserve` credentials, `pushNotifications.vapidPrivateKey`, SMTP credentials, and `authorizedEmails`.
- Swap the middleware to `requireAdmin`. The handler was already only consumed by `AdminPortal/*` pages (the existing comment read "Return full config for admin to edit"), so this is the privilege level it always meant to require.
- Did **not** strip/mask secrets in the response: `PUT /api/config` does a shallow merge of whatever the client posts back, so masking on GET would cause admin saves to silently wipe the stored secrets. The viewer/manager exposure is the actual bug; admin readability of secrets they need to edit is intentional.

## Test plan

- [ ] CI typecheck/build passes.
- [ ] As `viewer`/`manager`, `GET /api/config` now returns 403.
- [ ] As `admin`, `GET /api/config` still returns the full config and the Settings UI loads/saves without regression (OpenAI key, SMTP, push notifications, OAuth, OpenObserve sections).
- [ ] `AdminPortal` metrics-tab visibility check still works for admins; the silent failure on the same call for non-admins is acceptable (catch block already swallows it).

Closes ticket #534.

### DIFF
--- a/backend/src/routes/config.ts
+++ b/backend/src/routes/config.ts
@@ -51,6 +51,93 @@ const CONFIG_PATH = path.join(DATA_DIR, "config.json");
 const LOGS_DIR = path.join(DATA_DIR, "logs");
 
 /**
+ * Sentinel string returned in GET /api/config in place of any non-empty
+ * sensitive value. The PUT handler treats this exact string as "leave the
+ * stored value alone", restoring the on-disk value before persisting. This
+ * keeps plaintext secrets off the wire (and out of browser memory / dev
+ * tools) while preserving the admin-editor round-trip.
+ */
+const SECRET_PLACEHOLDER = "__SECRET_UNCHANGED__";
+
+/**
+ * Dot-notation paths to fields whose plaintext values must never appear
+ * in API responses. Paths that don't exist in a given config are silently
+ * skipped, so adding new sensitive fields here is safe.
+ */
+const SENSITIVE_CONFIG_PATHS: readonly string[] = [
+  "environment.auth.sessionSecret",
+  "environment.auth.google.clientSecret",
+  "analytics.openobserve.password",
+  "openai.apiKey",
+  "pushNotifications.vapidPrivateKey",
+  "email.smtp.auth.user",
+  "email.smtp.auth.pass",
+];
+
+function getDeep(obj: any, dottedPath: string): any {
+  const parts = dottedPath.split(".");
+  let cursor = obj;
+  for (const key of parts) {
+    if (cursor === null || cursor === undefined || typeof cursor !== "object") {
+      return undefined;
+    }
+    cursor = cursor[key];
+  }
+  return cursor;
+}
+
+function setDeep(obj: any, dottedPath: string, value: any): void {
+  const parts = dottedPath.split(".");
+  let cursor = obj;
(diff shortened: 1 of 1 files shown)


theodorecharles:

The core fix is solid: swapping `requireAuth` for `requireAdmin` on `GET /api/config` correctly gates the endpoint so viewers and managers receive a 403, satisfying criteria 1 and 3. The `requireAdmin` middleware enforces a strict `role === 'admin'` check and the import cleanup is safe (no other uses of `requireAuth` remain in the file). The unmet criterion is #2 — the ticket explicitly requires sensitive fields to be masked or absent even for admin callers, and the PR deliberately skips this. The author's rationale is technically valid (the shallow-merge `PUT` handler would corrupt stored secrets if the GET returned placeholders), but the right response is to fix the PUT strategy or get explicit ticket sign-off to relax the criterion — not to ship without it. Until that's resolved the PR doesn't fully close #534.

**Blockers**
- `backend/src/routes/config.ts:121` — Criterion 2 is unmet: the ticket requires sensitive fields (sessionSecret, clientSecret, apiKey, SMTP credentials, vapidPrivateKey) to be masked or absent in the response body even for admin callers. The PR intentionally skips this because the `PUT` handler does a shallow merge that would overwrite stored secrets with any placeholder values returned by GET. That concern is real, but the fix is to make the PUT handler skip sentinel values (e.g. re-read the current value from disk for any field equal to `'***'` or similar) rather than leave plaintext secrets flowing to the browser. Shipping without masking means the HMAC session key, OAuth client secret, OpenAI API key, and VAPID private key are still exposed to every admin session — a smaller blast radius than the original bug but still a material risk if any admin account is compromised.

theodorecharles:

The PR satisfies all three acceptance criteria. `requireAdmin` gates the GET endpoint and correctly returns 403 for viewer and manager roles. Despite the PR description claiming no masking was done, the code actually adds `maskSensitiveFields()` — covering all seven credential paths named in the ticket — and pairs it with `restoreSensitiveFields()` in the PUT handler so sentinel values are never written to disk. The round-trip design is sound. Only minor gap is that `authorizedEmails` (flagged sensitive in the ticket body) is absent from `SENSITIVE_CONFIG_PATHS`, but it's not named in the acceptance criteria, so this is a follow-on concern rather than a blocker.

**Suggestions**
- `backend/src/routes/config.ts:60` — `authorizedEmails` was listed as sensitive in the ticket body — worth adding to `SENSITIVE_CONFIG_PATHS` in a follow-up so the email allowlist isn't visible to admins who shouldn't see individual addresses (e.g. in a multi-admin setup).

**Nits**
- `backend/src/routes/config.ts:53` — The PR description says masking was *not* implemented, but the code clearly does implement it via `maskSensitiveFields`. The description should be corrected to avoid confusing future readers of the git log.

PR #219, opened 2026-04-27
fix: drop dead foreign_keys pragma toggles inside album rename txn (ticket #505)

theodorecharles:
## Summary

`PATCH /:album/rename` wrapped its UPDATEs in `db.transaction(...)` and toggled `db.pragma('foreign_keys = OFF')` / `... = ON'` around them. Per [SQLite docs](https://sqlite.org/pragma.html#pragma_foreign_keys), the `foreign_keys` pragma is a no-op while a transaction is active, so the toggles never did anything — FK enforcement state was unchanged.

The schema in `backend/src/database.ts:170` already declares `share_links.album` with `ON UPDATE CASCADE` (and the dedicated `migrate-share-links-cascade.js` migration brings older databases up to that schema). The cascade fires automatically on the album rename, so the toggles aren't needed.

Removed the two dead `db.pragma(...)` lines and rewrote the surrounding comment to explain what's actually happening. The explicit `UPDATE share_links ...` is left in place as a safety net for any database that hasn't run the cascade migration yet.

## Test plan

- [ ] CI build + tests pass on the PR
- [ ] Manual: rename an album that has a `share_links` row → `albums`, `image_metadata`, and `share_links` all reflect the new name; share link still resolves

### DIFF
--- a/backend/src/routes/album-management.ts
+++ b/backend/src/routes/album-management.ts
@@ -1168,39 +1168,35 @@ router.patch("/:album/rename", requireManager, async (req: Request, res: Respons
     // This way if DB update fails, filesystem is unchanged
     const db = getDatabase();
     
-    // Start transaction with foreign keys temporarily disabled
-    // This is needed because share_links has FK to albums(name) without ON UPDATE CASCADE
+    // share_links.album has ON UPDATE CASCADE (see database.ts), so renaming
+    // the album row cascades automatically. The explicit share_links UPDATE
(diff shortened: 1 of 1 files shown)


theodorecharles:

Both dead pragma calls are removed and nothing else is touched beyond an accurate comment rewrite and minor trailing-whitespace cleanup. The acceptance criteria are fully met: the `foreign_keys` toggles are gone and the album-rename UPDATE logic is otherwise identical. Clean, minimal fix.

PR #220, opened 2026-04-27
fix: require re-auth proof on MFA disable and backup-code reset (ticket #536)

theodorecharles:
## Summary

- `/api/auth-extended/mfa/disable` and `/api/auth-extended/mfa/backup-codes` guarded with `if (user.password_hash && !verifyPassword(user, password))`. When `password_hash` is null/empty (Google OAuth or passkey-only accounts), the `&&` short-circuits to `false` and the request goes through with no proof of identity — an attacker who steals a Google session can disable MFA or mint fresh backup codes, defeating the second factor.
- Replaces the half-check with a new `verifyReauth(user, body)` helper that accepts either a valid password (when the account has one) or a fresh passkey assertion bound to the user's own credential. Rejects when neither is supplied.
- Passkey path verifies the credential belongs to the currently authenticated user, consumes the challenge, and bumps the credential counter to prevent replay.

## Notes

- Backwards compatible for users with a password — the existing `{ password }` body keeps working unchanged on both endpoints.
- OAuth-only users with MFA enabled who go through today's UI (which only collects a password) will now get a 401 instead of silently bypassing the check. This is the intended secure behavior. A small frontend uplift to drive `/passkey/auth-options` and pass `passkeyCredential` + `passkeySessionId` to these endpoints is a follow-up — the API contract is in place.

## Test plan

- [ ] User with password: `POST /mfa/disable` with correct password → 200; with wrong/missing password → 401.
- [ ] User without password (OAuth-only) and no passkey body → 401 with re-auth-required message (was: 200, MFA disabled).
- [ ] User without password supplying valid passkey assertion for their own credential → 200.
- [ ] User supplying a passkey assertion for a credential belonging to a different user → 401.
- [ ] Same checks against `POST /mfa/backup-codes`.

### DIFF
--- a/backend/src/routes/auth-extended.ts
+++ b/backend/src/routes/auth-extended.ts
@@ -98,6 +98,78 @@ function getUserIdFromRequest(req: Request): number | null {
 // Store temporary challenges in memory (in production, use Redis or session store)
 const challenges = new Map<string, { challenge: string; userId?: number; user?: User; expires: number }>();
 
+/**
+ * Verify a sensitive-action re-authentication proof.
+ *
+ * Accepts either a valid password (when the user has a `password_hash`) or a
+ * fresh passkey assertion bound to the user's own credential. This closes a
+ * bypass where users without a password (Google OAuth / passkey-only) could
+ * pass through `if (user.password_hash && !verifyPassword(...))` because the
+ * `&&` short-circuits to `false` when `password_hash` is empty, allowing MFA
+ * to be disabled or backup codes to be regenerated with no proof.
+ */
+async function verifyReauth(
+  user: User,
+  body: any
+): Promise<{ ok: true } | { ok: false; status: number; reason: string }> {
+  const { password, passkeyCredential, passkeySessionId } = body || {};
+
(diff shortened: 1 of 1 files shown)


theodorecharles:

The fix is clean and directly addresses all three acceptance criteria. Both `/mfa/disable` and `/mfa/backup-codes` now route through `verifyReauth()`, which correctly handles the three cases: (1) a valid passkey assertion — cryptographically verified, cross-user ownership checked, challenge consumed and counter bumped; (2) a password when the account has one; (3) neither present → 401 regardless of `password_hash` state. Password-based users are entirely unaffected since Path 2 short-circuits before the fallthrough rejection. The ownership guard (`result.user.id !== user.id`) closes the obvious stolen-credential angle on the passkey path. The PR notes honestly that OAuth-only users hitting the current UI will start receiving 401s, which is the correct secure behavior pending a frontend follow-up. The in-memory `challenges` map is pre-existing infrastructure, not introduced here. No automated backend route tests exist anywhere in the repo, so the manual test plan is consistent with the codebase norm. Scope is one file, no unrelated edits.

PR #223, opened 2026-04-27
fix: gate /users with requireAdmin and stop leaking invite_token (ticket #535)

theodorecharles:
## Summary
Fixes a HIGH severity privilege escalation in `GET /api/auth-extended/users`: the route was gated by `requireAuth` instead of `requireAdmin`, so any authenticated user (including a viewer) could enumerate every account and read each invited user's `invite_token`. Those tokens are accepted by `GET /invite/:token` and `POST /invite/:token/complete`, so a viewer could claim an invited admin account before the legitimate invitee did.

- `GET /api/auth-extended/users` now requires admin and no longer returns `invite_token` in any list row.
- New admin-only `GET /api/auth-extended/users/:userId/invite-link` returns the invite URL (and token) for a single invited user, so the admin "Copy Invite Link" button keeps working without bulk-leaking tokens.
- Frontend `UserCard` now fetches the URL on demand via the new endpoint instead of reading `invite_token` off each list row, and the `User` TS type drops the now-removed `invite_token` field.

## Test plan
- [ ] As a viewer: `GET /api/auth-extended/users` returns 403.
- [ ] As an admin: `GET /api/auth-extended/users` returns 200, response rows have no `invite_token` field.
- [ ] As an admin: clicking "Copy Invite Link" on an invited user copies the correct `/invite/<token>` URL.
- [ ] As an admin: `GET /api/auth-extended/users/:userId/invite-link` returns 400 for active users and 200 for invited users.
- [ ] As a viewer: `GET /api/auth-extended/users/:userId/invite-link` returns 403.

### DIFF
--- a/backend/src/routes/auth-extended.ts
+++ b/backend/src/routes/auth-extended.ts
@@ -1429,17 +1429,21 @@ router.delete('/passkey/:id', requireAuth, async (req: Request, res: Response) =
 
 /**
  * List all users (admin only)
+ *
+ * Security note: requires `requireAdmin` — non-admin accounts must not be able
+ * to enumerate users. The response intentionally omits `invite_token`; admins
+ * who need an outstanding invite URL fetch it via
+ * `GET /users/:userId/invite-link` (one user at a time, admin-gated).
  */
-router.get('/users', requireAuth, (req: Request, res: Response) => {
+router.get('/users', requireAdmin, (req: Request, res: Response) => {
   try {
     const users = getAllUsers();
-    
+
     // Remove sensitive data and add computed fields
     const sanitizedUsers = users.map((user: User) => {
       // Determine display status based on user state
       let displayStatus = null;
-      let inviteToken = null;
-      
+
       if (user.status === 'invited') {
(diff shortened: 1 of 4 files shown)


theodorecharles:

All three acceptance criteria are met: the list-users route now requires `requireAdmin` (not `requireAuth`), `invite_token` is removed from every code path in the map — both the variable and the returned field are gone — and admins can still reach the endpoint. The new `/users/:userId/invite-link` admin-only route is necessary to preserve the Copy Invite Link UX and is properly gated. Frontend updates (type drop, on-demand fetch, button condition) are all consequential. No existing tests to benchmark against. The new endpoint redundantly returns `inviteToken` alongside `inviteUrl`, but since it's admin-only and single-user this is a nit, not a concern.

**Nits**
- `backend/src/routes/auth-extended.ts:1510` — The response includes `inviteToken` in addition to `inviteUrl`, but the frontend only reads `inviteUrl`. Returning the raw token is harmless given this endpoint is admin-only, but it could be dropped to keep the surface minimal.

PR #224, opened 2026-04-27
fix: rename album filesystem before DB to avoid inconsistent state (ticket #504)

theodorecharles:
## Summary

- `PATCH /:album/rename` was committing the DB transaction before any `fs.renameSync`, so a filesystem rename failure (EBUSY / EACCES / ENOSPC / race) left the DB pointing at the new name while disk still had the old name — every subsequent read hit ENOENT with no rollback.
- Swapped the order to mirror the sibling `PUT /:album/rename`: filesystem first, then DB. Each fs rename is tracked so a partial failure rolls back prior renames; a DB transaction failure also rolls back every fs rename, leaving the album in its original on-disk state.
- Closes ticket #504.

## Test plan

- [ ] CI typecheck/build passes on the backend.
- [ ] Manual: rename an album via `PATCH /:album/rename` — DB and on-disk paths both move to the new name.
- [ ] Manual: simulate a fs rename failure (e.g. take a lock on `optimized/thumbnail/<album>` mid-rename) and confirm the photos directory and DB are reverted to the old name on 500.
- [ ] Manual: simulate a DB constraint failure during the transaction — confirm photos / optimized / video dirs are renamed back to the old name on 500.

### DIFF
--- a/backend/src/routes/album-management.ts
+++ b/backend/src/routes/album-management.ts
@@ -1164,72 +1164,107 @@ router.patch("/:album/rename", requireManager, async (req: Request, res: Respons
       return;
     }
 
-    // Update database FIRST before touching filesystem
-    // This way if DB update fails, filesystem is unchanged
+    // Rename filesystem directories FIRST, tracking each successful rename
+    // so we can roll back on a later filesystem or database failure. This
+    // keeps the album consistent across DB and disk: either every rename
+    // sticks or none of them do.
+    const renamedPaths: Array<{ from: string; to: string }> = [];
+    const rollbackFsRenames = () => {
+      for (const { from, to } of [...renamedPaths].reverse()) {
+        try {
+          if (fs.existsSync(to)) {
+            fs.renameSync(to, from);
+          }
+        } catch (rollbackErr) {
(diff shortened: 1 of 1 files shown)


theodorecharles:

The PR cleanly satisfies all three acceptance criteria. Filesystem renames now happen before any DB work: each successful rename is pushed onto `renamedPaths` so the `rollbackFsRenames` helper can reverse them in order if a later rename or the DB transaction throws. The two `try/catch` blocks correctly cover both failure modes — fs failure before the transaction is ever called, and DB failure after all fs renames have landed. The implementation mirrors the safer `PUT /:album/rename` pattern the ticket pointed to, the scope is one file and one handler, and there's no unrelated churn beyond whitespace normalization inside the transaction body.

**Suggestions**
- `backend/src/routes/album-management.ts:1206` — Inside the `transaction()` callback, `db.pragma('foreign_keys = OFF')` is set at the start and `ON` at the end. SQLite pragmas are connection-level, not transactional — if the callback throws (e.g. "Album not found in database") and better-sqlite3 rolls back the data changes, the pragma stays `OFF` on the connection. This is a pre-existing issue from before this PR, not introduced here, but it means any code that runs on the same connection immediately after a failed rename could skip FK enforcement. Worth a follow-up ticket rather than blocking this fix.

PR #225, opened 2026-04-27
fix(auth): require admin to list users so invite tokens stop leaking (ticket #32)

theodorecharles:
## Summary

`GET /api/auth-extended/users` was gated by `requireAuth` but returned the raw `invite_token` for every user in the `invited` state. `POST /api/auth-extended/invite/:token/complete` is unauthenticated by design and lets the caller pick a name+password and activate the account, so any low-privilege authenticated user (viewer/manager) could list pending invitations and seize an invited admin/manager account — a direct privilege-escalation path.

This change switches the endpoint to `requireAdmin`, matching every sibling user-management route (`POST /invite`, `POST /invite/resend/:userId`, `DELETE /users/:userId`, `PATCH /users/:userId/role`, `POST /users/:userId/reset-mfa`, `POST /users/:userId/send-password-reset`) and the route's own `(admin only)` JSDoc. An inline comment now explains why the admin gating must stay in place while the response continues to include `invite_token` for the admin "Copy Invite Link" UI.

## Frontend impact

None for legitimate users. The user-management UI is only rendered to `currentUser.role === 'admin'`. Non-admins who hit the endpoint now get `403` instead of `200`-with-sensitive-data.

## Test plan

- [ ] CI passes on the PR (no backend tests exist locally to run)
- [ ] As a `viewer`, `GET /api/auth-extended/users` returns `403 Access denied - admin role required`
- [ ] As a `manager`, `GET /api/auth-extended/users` returns `403`
- [ ] As an `admin`, `GET /api/auth-extended/users` still returns the user list with `invite_token` populated for invited users
- [ ] Admin "Copy Invite Link" button in the user-management UI still works end-to-end

### DIFF
--- a/backend/src/routes/auth-extended.ts
+++ b/backend/src/routes/auth-extended.ts
@@ -1430,24 +1430,28 @@ router.delete('/passkey/:id', requireAuth, async (req: Request, res: Response) =
 /**
  * List all users (admin only)
  *
- * SECURITY: This endpoint returns `invite_token` for users in the `invited`
- * state so the admin UI can render a "Copy Invite Link" affordance. The
+ * SECURITY: This endpoint MUST NOT return `invite_token` for any user. The
  * invite token is sufficient on its own to complete the invitation and take
  * over the invited account (see POST /invite/:token/complete, which is
- * unauthenticated by design). It MUST therefore stay gated by requireAdmin —
- * downgrading this to requireAuth leaks invite tokens to viewers/managers and
- * creates a direct privilege-escalation path.
+ * unauthenticated by design), so disclosing it on every page-load of the
+ * admin user list — even to admins — needlessly widens the blast radius of a
+ * compromised admin session and blocks future token-rotation / one-time-use
+ * controls. Admins fetch the token on demand via
+ * GET /users/:userId/invite-link when they explicitly click "Copy Link".
+ *
+ * Endpoint also stays gated by requireAdmin — downgrading to requireAuth
+ * would expose user metadata (roles, auth methods, MFA state, etc.) to
+ * viewers/managers.
  */
 router.get('/users', requireAdmin, (req: Request, res: Response) => {
   try {
     const users = getAllUsers();
-    
+
     // Remove sensitive data and add computed fields
     const sanitizedUsers = users.map((user: User) => {
       // Determine display status based on user state
       let displayStatus = null;
-      let inviteToken = null;
-      
+
       if (user.status === 'invited') {
         // Check if invite has expired
         if (user.invite_expires_at) {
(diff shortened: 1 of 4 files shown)


theodorecharles:

The PR correctly upgrades the `GET /api/auth-extended/users` middleware from `requireAuth` to `requireAdmin`, which satisfies AC1 (non-admins now get 403). The security comment is a nice addition. However, it does not satisfy AC2: `invite_token` is still present in every invited/expired-invite user record returned to the caller, just now only admin callers. The acceptance criterion reads 'invite_token is not present in the response body of the user-listing endpoint for any role' — 'any role' is unambiguous and includes admin. The ticket body explicitly anticipated this case and proposed a separate admin-only endpoint for generating/fetching the copyable link. The risk of leaving the raw token in a list response even under admin-only gating is real: it widens the blast radius if an admin session is compromised, it bypasses any future token-rotation or one-time-use controls, and it means the token is served to every admin on every page-load of the user list rather than on explicit 'Copy Link' intent. Until `invite_token` is stripped from the list response and the admin Copy-Invite-Link UI is wired to a dedicated endpoint, AC2 is unmet and the original vulnerability surface is only partially closed.

**Blockers**
- `backend/src/routes/auth-extended.ts:1470` — The line `invite_token: inviteToken` still serializes the raw invite token into every user-list response for invited/expired accounts, now visible to every admin. AC2 requires the token to be absent from this endpoint's response for *any* role. Strip the field here and expose it (or a freshly-generated invite URL) through a dedicated `GET /api/auth-extended/users/:userId/invite-link` route that is separately gated by `requireAdmin`. That limits token exposure to the exact moment an admin intentionally clicks 'Copy Link', makes future token rotation and one-time-use semantics straightforward to add, and puts the correct minimal-disclosure shape on the list response.

theodorecharles:

Both acceptance criteria are fully met. `GET /users` is now gated by `requireAdmin`, which returns `403 Access denied - admin role required` for any non-admin authenticated session, and `invite_token` is gone from every list response regardless of caller role. The companion `GET /users/:userId/invite-link` endpoint is a necessary addition — without it the admin "Copy Invite Link" button would simply break — and it's correctly gated by `requireAdmin` as well. The fix matches every other user-management route's auth posture and the route's own JSDoc. Comment trail is unusually clear. One edge-case worth addressing before it bites: the new invite-link handler guards with `user.status !== 'invited'`, but `'invite_expired'` is a distinct raw DB status (visible in `database-users.ts` and the resend-invite guard at line 426). In practice this path is probably unreachable today because the list endpoint also only sets `displayStatus` for `user.status === 'invited'`, so the button never renders for a raw-`invite_expired` user — but the guard is fragile. Not a blocker for the security fix.

**Suggestions**
- `backend/src/routes/auth-extended.ts:1503` — The guard `user.status !== 'invited'` returns 404 for any user whose raw DB status is `'invite_expired'` (a real DB value per `database-users.ts`). Today the UI never renders "Copy Invite" for such users, so it's not reachable — but a future change could expose it. Widening this to `!['invited', 'invite_expired'].includes(user.status)` would make the endpoint self-consistent with the resend-invite guard at line 426 and resilient to future list-response changes.

PR #226, opened 2026-04-27
Fix public EXIF endpoint leaking GPS coordinates (ticket #35)

theodorecharles:
## Summary

`GET /api/photos/:album/:filename/exif` was mounted with no auth gate and called `exifr.parse(filePath)` with default options, returning the full EXIF object — including `GPSLatitude`/`GPSLongitude`/`GPSAltitude` — to any visitor who knew (or could enumerate) a filename. This contradicts the documented privacy guarantee that public album variants strip GPS, and the published-state of the album wasn't even checked.

This change in `backend/src/routes/albums.ts`:

- Gates `servePhotoEXIF` on the same `isAuthenticated` + `albumState.published` check used by `serveAlbumPhotos`: 404 for missing albums, 403 for unpublished albums when the caller is not authenticated.
- For unauthenticated callers, parses with `{ gps: false }` so `exifr` never extracts GPS coordinates, and defensively strips any residual `GPS*` / `latitude` / `longitude` / `altitude` keys (defense-in-depth in case GPS leaks through a different block such as XMP).
- Authenticated admin callers continue to receive the full EXIF payload.

Closes ticket #35.

## Test plan

- [ ] CI passes on this PR.
- [ ] Hit `GET /api/photos/<published-album>/<file>/exif` unauthenticated → response contains EXIF but no `GPS*` / `latitude` / `longitude` / `altitude` keys.
- [ ] Hit the same endpoint authenticated as admin → full EXIF including GPS returned.
- [ ] Hit the endpoint for an unpublished album unauthenticated → 403.
- [ ] Hit the endpoint for an unknown album → 404.

### DIFF
--- a/backend/src/routes/albums.ts
+++ b/backend/src/routes/albums.ts
@@ -571,7 +571,7 @@ const servePhotoEXIF = async (req: Request, res: Response, albumName: string, fi
   // Sanitize inputs to prevent path traversal
   const sanitizedAlbum = sanitizePath(albumName);
   const sanitizedFilename = filename.replace(/[^a-zA-Z0-9_. -]/g, '');
-  
+
   if (!sanitizedAlbum || !sanitizedFilename) {
     res.status(400).json({ error: "Invalid album or filename" });
     return;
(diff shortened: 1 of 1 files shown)


theodorecharles:

The PR correctly fixes the GPS-leaking EXIF endpoint. It mirrors the exact `isAuthenticated` + `getAlbumState` pattern already used by `serveAlbumPhotos` and `serveAlbumJSON` — same two-part auth check (`req.isAuthenticated()` || `session.userId`), same 404/403 ordering, same function. For unauthenticated callers it passes `{ gps: false }` to exifr (suppresses the GPS IFD block at parse time) and then does a second-pass key-delete sweep for `GPS*`, `latitude`, `longitude`, and `altitude` to catch anything that might leak via XMP or IPTC. Authenticated callers continue to receive the full payload unchanged. Both acceptance criteria are satisfied: GPS coordinates cannot reach an unauthenticated caller, and album publication state is gated before any EXIF data is served. No scope creep, no unrelated changes, and the approach fits the codebase style. The share-link edge case (an unpublished album accessed via a valid secret link) would now return 403 on the EXIF sub-path rather than GPS-stripped EXIF, but that gap pre-dated this PR and is not in scope for this ticket — and it's the safer side of the line.

PR #228, opened 2026-04-27
fix: harden /api/system/restart with CSRF + per-IP rate limit (ticket #634)

theodorecharles:
## Summary

Closes ticket #634 — `[security-audit] Authenticated viewer can restart the server (DoS) via /api/system/restart`.

The endpoint already required `requireAdmin` (the audit finding's evidence was stale on this point), but had no CSRF check and no endpoint-specific rate limit. Two changes in `backend/src/routes/system.ts`:

- Mount `csrfProtection` on the system router (same pattern used in `routes/config.ts` and `routes/branding.ts` — middleware no-ops on safe methods, enforces auth + same-origin on POST/PUT/DELETE).
- Add a strict per-IP `express-rate-limit` of 3 requests per 15 minutes specifically for `/restart`. The global `/api` limiter is too permissive for an action that takes the process down for every other user; a stolen admin session or a buggy client can no longer loop the endpoint into a denial-of-service.

`requireAdmin` was already in place, so no auth-level changes were needed.

## Test plan

- [ ] CI: backend `tsc` build succeeds.
- [ ] Manual: as an admin, hit `POST /api/system/restart` four times in quick succession — fourth call returns 429.
- [ ] Manual: a cross-origin POST (no/invalid Origin header) returns 403 from `csrfProtection`.
- [ ] Manual: same-origin admin restart still works as before.

### DIFF
--- a/backend/src/routes/system.ts
+++ b/backend/src/routes/system.ts
@@ -8,7 +8,9 @@ import { spawn } from 'child_process';
 import path from 'path';
 import fs from 'fs';
 import { fileURLToPath } from 'url';
+import rateLimit from 'express-rate-limit';
 import { requireAdmin } from '../auth/middleware.js';
+import { csrfProtection } from '../security.js';
 import { error, warn, info, debug, verbose } from '../utils/logger.js';
 
 const __filename = fileURLToPath(import.meta.url);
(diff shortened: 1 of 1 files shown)


theodorecharles:

The PR satisfies all four acceptance criteria. `csrfProtection` is mounted on the router via `router.use(csrfProtection)` — matching the pattern already in `config.ts` and `branding.ts` — and the `restartLimiter` (3 req / 15 min, per-IP) is correctly wired in front of the handler. `requireAdmin` was already in place on the PR's base branch, which the PR description explains accurately. `express-rate-limit` is already a declared dependency so there's nothing new to install. One ordering nit worth reading: the rate limiter is applied before `requireAdmin`, which means a non-admin (or unauthenticated) user sharing an IP with an admin could exhaust the three-shot allowance before a legitimate admin restart lands. The impact is bounded — an exhausted limit delays but does not permanently block the admin, and the window is short — but swapping the order to `requireAdmin, restartLimiter` would tighten this. The `csrfProtection` function's no-origin-no-referer pass-through is a pre-existing gap in the project's CSRF utility, not introduced here, so I'm not flagging it as a blocker on this PR.

**Suggestions**
- `backend/src/routes/system.ts:41` — The rate limiter fires before `requireAdmin`, so a non-admin user on the same IP can exhaust the three-request window and delay a legitimate admin restart. Consider reordering to `requireAdmin, restartLimiter` so the per-IP counter only ticks for requests that have already passed the admin gate.

PR #229, opened 2026-04-27
fix: gate preview-grid album route on published state and sanitize name (ticket #629)

theodorecharles:
## Summary

`GET /api/preview-grid/album/:albumName` previously rendered a 2x2 social-preview grid for any directory under `photosDir`, with no auth, no share-link check, and no album-state lookup. The param was also interpolated into `path.join` without going through the project-wide `sanitizeName`. The `/homepage` and `/shared/:secretKey` siblings already gate access via `getImagesForHomepage` and the share-link checks — this brings `/album` in line.

- Added a local `sanitizeName` helper matching `share-links.ts` / `album-management.ts` and applied it to the `:albumName` param. Invalid names 404 before any DB or filesystem work.
- Imported `getAlbumState` from `database.js` and 404 the route when the album is missing or `published === false`.
- Use the sanitized name everywhere downstream (`path.join`, `generatePhotoGrid`, ETag) so the raw param never reaches disk.

## Test plan

- [ ] CI: backend type-check / lint pass on the PR
- [ ] Manual: `curl /api/preview-grid/album/<published-album>` still returns the JPEG grid
- [ ] Manual: `curl /api/preview-grid/album/<unpublished-album>` returns 404
- [ ] Manual: `curl /api/preview-grid/album/..%2Fetc%2Fpasswd` returns 404 (sanitizer rejects)
- [ ] Manual: `/homepage` and `/shared/:secretKey` previews unchanged

Closes ticket #629.

### DIFF
--- a/backend/src/routes/preview-grid.ts
+++ b/backend/src/routes/preview-grid.ts
@@ -7,11 +7,25 @@ import { Router, Request, Response } from "express";
 import sharp from "sharp";
 import fs from "fs";
 import path from "path";
-import { getShareLinkBySecret, isShareLinkExpired, getImagesForHomepage } from "../database.js";
+import { getShareLinkBySecret, isShareLinkExpired, getImagesForHomepage, getAlbumState } from "../database.js";
 import { DATA_DIR } from "../config.js";
 import { error, warn, info, debug, verbose } from '../utils/logger.js';
 
 const router = Router();
+
+/**
+ * Sanitize album name - mirrors the shared sanitizer used by album/share-link routes.
+ * Allows letters, numbers, spaces, hyphens, and underscores.
+ */
+const sanitizeName = (name: string): string | null => {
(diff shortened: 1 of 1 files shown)


theodorecharles:

The diff satisfies all three acceptance criteria cleanly. `sanitizeName` is an exact copy of the function from `share-links.ts` and `album-management.ts`, applied to `:albumName` before any filesystem or DB work — the raw param never reaches `path.join`, `generatePhotoGrid`, or the ETag header. The `getAlbumState` call is synchronous (consistent with every other use in the codebase, and the function explicitly coerces `published` to a boolean), and the route returns 404 for both missing and unpublished albums. Published albums remain accessible to everyone without any new auth gate, satisfying the third criterion. The diff is tightly scoped to one file — the whitespace normalization is cosmetic and harmless. No automated tests exist for any preview-grid route, so the absence of a new test file isn't a regression. CI checks weren't reported on the branch yet, but the change is straightforward enough that I have no concern there.

PR #232, opened 2026-04-27
fix: close TOCTOU race in verifyBackupCode (ticket #624)

theodorecharles:
## Summary

`verifyBackupCode` (in `backend/src/database-users.ts`) read the caller's stale `user.backup_codes` snapshot, bcrypt-matched, then UPDATEd with a filtered list. Two concurrent MFA verifications presenting the same backup code could both match and both write back the same filtered list — the code was consumed once in the DB but accepted twice (or N times). Backup codes are explicitly single-use, so this defeated that guarantee.

The verify/consume sequence now runs inside a `BEGIN IMMEDIATE` better-sqlite3 transaction:

- The row is re-read **inside** the transaction so we match against canonical state, not the caller's snapshot.
- The UPDATE is conditional on the old JSON: `WHERE id = ? AND backup_codes = ?`. If a concurrent writer committed first, the conditional UPDATE affects 0 rows and we return `false` (fail closed).
- `BEGIN IMMEDIATE` acquires a RESERVED lock at transaction start, serializing writers across processes (PM2 cluster mode under WAL).

Function signature is unchanged, so the route handler in `auth-extended.ts` did not need edits.

## Test plan

- [ ] CI build passes (`tsc`).
- [ ] Manual: enable MFA, generate backup codes, log out, verify a backup code works exactly once and is rejected on subsequent attempts.
- [ ] Optional repro: fire two concurrent `/api/auth/login` requests with the same backup code in `mfaToken` against a multi-process deployment; only one should succeed.

### DIFF
--- a/backend/src/database-users.ts
+++ b/backend/src/database-users.ts
@@ -282,36 +282,71 @@ export function disableMFA(userId: number): boolean {
 }
 
 /**
- * Verify backup code and remove it if valid
+ * Verify backup code and remove it if valid.
+ *
+ * Backup codes are single-use. To prevent a TOCTOU race where two concurrent
+ * MFA verifications presenting the same code both read the original list,
+ * both bcrypt-match, and both write back the same filtered list (consuming
+ * the code once in the DB but accepting it N times), the read/match/update
+ * is performed inside a `BEGIN IMMEDIATE` transaction. The row is re-read
+ * inside the transaction, and the UPDATE is conditional on the old JSON
+ * snapshot so a concurrent writer that committed first causes us to fail
+ * closed.
  */
 export function verifyBackupCode(user: User, code: string): boolean {
   if (!user.backup_codes || user.backup_codes.length === 0) {
     return false;
   }
-  
-  // Find matching backup code
-  let matchIndex = -1;
-  for (let i = 0; i < user.backup_codes.length; i++) {
-    if (bcrypt.compareSync(code, user.backup_codes[i])) {
-      matchIndex = i;
-      break;
-    }
-  }
-  
-  if (matchIndex === -1) return false;
-  
-  // Remove used backup code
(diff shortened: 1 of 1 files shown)


theodorecharles:

The fix correctly closes the TOCTOU window. The read/match/update sequence is now inside a `db.transaction()` called with `.immediate()`, which acquires a RESERVED lock at transaction start and serializes writers. The row is re-read fresh inside the transaction (discarding the stale `user` snapshot), bcrypt comparison runs against canonical state, and the UPDATE is further guarded by `WHERE backup_codes = ?` so a concurrent writer that commits first will cause this to affect 0 rows and return `false`. All three acceptance criteria are met: one success and one failure on concurrent same-code attempts, read-match-invalidate wrapped atomically, and a consumed code cannot be reused. The diff is tightly scoped to `verifyBackupCode` with no unrelated changes, and the function signature is unchanged so the call site in `auth-extended.ts` needed no edits. The one thing worth flagging — not a blocker — is that `bcrypt.compareSync` runs inside the transaction, holding the RESERVED lock for the full bcrypt wall-clock time (~100ms at cost 10). In a high-concurrency deployment this serializes all backup-code verifications through a single lock window; a future optimization would be to pre-screen candidates outside the transaction and re-verify inside, but that adds complexity and this form is correct and safe today.

**Suggestions**
- `backend/src/database-users.ts:322` — `bcrypt.compareSync` runs inside the `BEGIN IMMEDIATE` transaction, holding the RESERVED lock for the full bcrypt wall-clock time (~100ms at default cost). Readers are unaffected, but all concurrent backup-code writers are serialized through this window. Acceptable today; if throughput becomes a concern, consider scanning for a candidate hash outside the transaction and re-confirming inside before the conditional UPDATE.

PR #233, opened 2026-04-27
fix(security): sanitize branding.customCSS to prevent stored XSS (#633)

theodorecharles:
## Summary

Closes ticket #633 (Argus security audit, HIGH severity).

The frontend SSR layer was interpolating `branding.customCSS` directly into a `<style id="custom-css">…</style>` block on every served page, with no escaping or `</style>` filter. The backend `PUT /api/branding` endpoint accepted `customCSS` from any **manager** (a non-admin role) and validated only the 50KB length cap. A manager could ship something like `</style><script>fetch("/api/config")…</script>` and execute arbitrary JS in every visitor's browser, including admins — a clean stepping-stone to chained config/secret exfil.

This PR closes the breakout in two layers:

- **Backend** (`backend/src/routes/branding.ts`) — `PUT /api/branding` now rejects any `customCSS` containing `</`, `<!`, `<script`, or HTML-entity-encoded equivalents, returning `400` with a clear error. This is the authoritative gate.
- **Frontend defense-in-depth** (`frontend/server.js`, `scripts/generate-homepage-html.js`) — every customCSS injection site now runs the value through a `sanitizeCustomCSS()` helper that strips the same sequences. This protects against any pre-existing bad data already on disk in `config.json`.

`>` is intentionally still allowed because CSS uses it for child combinators (`.parent > .child`). The only sequences that can close a `<style>` raw-text element start with `<`, so blocking those is sufficient.

## Test plan

- [ ] CI build & typecheck pass (`backend` TS compiles with the new validator)
- [ ] Manually attempt `PUT /api/branding` with `{"customCSS": "</style><script>alert(1)</script>"}` → expect `400 Bad Request`
- [ ] Manually attempt `PUT /api/branding` with `{"customCSS": ".parent > .child { color: red; }"}` → expect `200 OK` (child combinators still work)
- [ ] Verify a page served by `frontend/server.js` with a clean `customCSS` still renders the `<style id="custom-css">` block correctly
- [ ] Confirm a config.json with a pre-existing malicious `customCSS` no longer emits `</style>...` into served HTML

### DIFF
--- a/backend/src/routes/branding.ts
+++ b/backend/src/routes/branding.ts
@@ -261,6 +261,25 @@ router.put("/", requireManager, (req: Request, res: Response) => {
               .json({ error: "Custom CSS is too long (max 50KB)" });
             return;
           }
+          // Reject sequences that could break out of the <style> raw-text
+          // element or smuggle in script content when this string is
+          // server-side interpolated into <style id="custom-css">…</style>
+          // by frontend/server.js. Blocks </style, </script, <!--, <script
+          // and HTML-entity-encoded equivalents (defense-in-depth — entities
+          // are not decoded inside <style>, but the audit asked for them).
+          // Note: we deliberately allow ">" since CSS uses it for child
+          // combinators (e.g. ".a > .b").
+          const dangerousCSS = /<\s*\/|<\s*!|<\s*script|&lt;\s*\/|&lt;\s*!|&lt;\s*script|&#0*60;\s*\/|&#0*60;\s*!|&#0*60;\s*script/i;
+          if (dangerousCSS.test(value)) {
+            warn(
+              "[Branding] Rejected customCSS containing HTML-tag-like sequences",
+            );
+            res.status(400).json({
+              error:
+                "Custom CSS may not contain HTML tags, comments, or closing-tag sequences (e.g. </, <!, <script). Use only CSS rules.",
+            });
+            return;
+          }
         } else if (value.length > 500) {
           res
             .status(400)
--- a/frontend/server.js
+++ b/frontend/server.js
@@ -429,6 +429,28 @@ function escapeHtml(str) {
 }
 
 /**
+ * Strip sequences from user-supplied CSS that could break out of a <style>
+ * raw-text element or smuggle in script content. Matches the same patterns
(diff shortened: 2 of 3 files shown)


theodorecharles:

The fix is correct and addresses all three acceptance criteria. The backend validator authoritatively rejects bad `customCSS` with a 400, and the frontend `sanitizeCustomCSS` helper strips the same sequences at every one of the four injection sites identified in the ticket plus the homepage-generation script — good defense-in-depth against stale `config.json` payloads. The regex is sound for the documented threat: `<\s*\/` covers the only real HTML-raw-text breakout vector, `<\s*script` catches tag openers, and the entity-encoded variants are belt-and-braces. No blockers, but two things worth noting before this ages.

**Suggestions**
- `frontend/server.js:449` — The `test`-then-`replace` pattern on a regex defined with the `g` flag leaves `lastIndex` advanced after the `test` call. `String.prototype.replace` does reset `lastIndex` to 0 for global regexps per spec, so the replacement itself is correct — but the idiom is fragile and confusing to future readers. A simpler and safer pattern is to skip the `test` entirely and just return `css.replace(dangerous, "")` unconditionally; `replace` returns the original string unchanged when there are no matches, so the early-return optimization isn't needed here. Same issue in `scripts/generate-homepage-html.js`.
- `backend/src/routes/branding.ts:271` — No automated test covers the new validator branch. Given the HIGH severity of this issue a unit test that submits `</style><script>alert(1)</script>` and asserts a 400 response — and a counterpart that submits `.parent > .child { color: red; }` and asserts 200 — would lock in the behaviour and catch any future regex regression. The PR's manual test plan is fine for the immediate merge, but the test gap is worth a follow-up ticket if adding it here isn't practical.

**Nits**
- `frontend/server.js:438` — The `sanitizeCustomCSS` function is copy-pasted verbatim between `frontend/server.js` and `scripts/generate-homepage-html.js`. If the regex ever needs updating (e.g. to cover `&#x3c;` hex-encoded variants) it has to be changed in two places. Extracting it to a small shared utility (e.g. `shared/sanitize-css.js`) would make future audits simpler — not a merge blocker, but worth a follow-up.

PR #234, opened 2026-04-27
fix(auth): close WebAuthn passkey-counter race (ticket #625)

theodorecharles:
## Summary

Closes the WebAuthn replay window in `updatePasskeyCounter` and the related TOCTOU bugs in `addPasskey` / `removePasskey` (ticket #625).

The three functions in `backend/src/database-users.ts` did a read-modify-write across separate statements: `getUserById` → mutate JSON in memory → `UPDATE users SET passkeys = ?`. Because the passkey verify path `await`s `verifyPasskeyAuthentication` between reading the user and calling `updatePasskeyCounter`, two concurrent assertions could both verify against the same stored counter and both write `N+1`. That defeats the WebAuthn signature counter — the whole point of which is to detect cloned authenticators / replayed assertions.

Fix:

- All three functions are wrapped in `db.transaction(...)`. better-sqlite3 is sync and single-connection, so the transaction body runs as one uninterruptible synchronous block — no `await` can sneak in between read and write.
- `updatePasskeyCounter` gains an `expectedCurrentCounter` arg and now does a compare-and-swap: if the stored counter has already moved past what the caller verified against, the update returns `false` (replay/concurrent-winner detected). It also defensively rejects writes that don't strictly advance.
- Both call sites in `backend/src/routes/auth-extended.ts` (re-auth and passkey login) pass the counter they verified against and treat a `false` return as authentication failure with a `warn` log.

No API surface changes outside the internal `updatePasskeyCounter` signature; both callers are in this repo and updated.

## Test plan

- [ ] CI passes on the PR (TypeScript build + tests)
- [ ] Manual: register a passkey, log in with it, confirm counter advances and login succeeds
- [ ] Manual: replay a captured passkey assertion — second use should fail with 401
- [ ] Manual: re-auth flow (`verifyReauth`) still accepts a fresh passkey assertion
- [ ] Manual: add and remove passkeys still work end-to-end

### DIFF
--- a/backend/src/database-users.ts
+++ b/backend/src/database-users.ts
@@ -316,84 +316,142 @@ export function verifyBackupCode(user: User, code: string): boolean {
 
 /**
  * Add passkey to user
+ *
+ * Wrapped in a SQLite transaction so the read-modify-write of the
+ * passkeys JSON blob is atomic. Without the transaction, two concurrent
+ * adds could each read the same row, append their own passkey, and the
+ * later write would clobber the earlier one (silent passkey loss).
  */
 export function addPasskey(userId: number, passkey: Omit<Passkey, 'created_at'>): boolean {
   const db = getDatabase();
-  const user = getUserById(userId);
-  if (!user) return false;
-  
-  const passkeys = user.passkeys || [];
-  passkeys.push({
-    ...passkey,
-    created_at: new Date().toISOString(),
-  });
-  
-  // Add 'passkey' to auth methods if not already there
-  const authMethods = [...user.auth_methods];
-  if (!authMethods.includes('passkey')) {
-    authMethods.push('passkey');
-  }
-  
+
   const stmt = db.prepare(`
-    UPDATE users 
+    UPDATE users
     SET passkeys = ?, auth_methods = ?, updated_at = CURRENT_TIMESTAMP
     WHERE id = ?
   `);
-  
-  const result = stmt.run(JSON.stringify(passkeys), JSON.stringify(authMethods), userId);
-  return result.changes > 0;
+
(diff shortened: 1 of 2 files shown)


theodorecharles:

The fix correctly closes all three paths described in the ticket. Wrapping each function in `db.transaction()` makes the read-modify-write atomic at the Node.js event-loop level (better-sqlite3 is fully synchronous, so no event can interleave with a transaction body). The compare-and-swap on `expectedCurrentCounter` provides a second layer of protection for the race window that opens during the async `verifyPasskeyAuthentication` call: whichever assertion wins the DB serialization gets the counter advanced, and the loser gets `false` and a 401. Both call sites in `auth-extended.ts` are updated correctly, passing the pre-verification counter snapshot as `expectedCurrentCounter`. `addPasskey` and `removePasskey` are similarly transaction-wrapped. No scope creep, no unrelated refactors, and the logic is idiomatic better-sqlite3.

**Suggestions**
- `backend/src/routes/auth-extended.ts:1404` — When `counterAdvanced` is false, the challenge is not deleted from the in-memory map and will linger until the TTL cleanup fires. For the replay-detection path that's fine, but a comment noting the intentional choice ("leave the challenge to expire naturally; do not delete so a retried request also fails") would make the reasoning obvious to the next reader.

**Nits**
- `backend/src/database-users.ts:452` — The zero-counter bypass (`if (expected !== 0 && next <= expected) return false`) is WebAuthn-spec-correct but silently disables replay detection for authenticators that never increment their counter. The comment above it mentions the carve-out, but it's worth a one-liner noting the implication: for those devices, replay protection falls entirely on the challenge rather than the counter.

PR #235, opened 2026-04-27
feat: show photo position counter in lightbox modal (ticket #620)

theodorecharles:
## Summary

- Renders a position counter (`14 / 47`) between the prev/next buttons in the lightbox `ModalNavigation` so viewers know where they are in large albums.
- Wires up the `currentIndex` / `totalPhotos` props in `ContentModal` that were already plumbed through and sitting idle behind a "kept for future UI features" comment.
- Hidden automatically when `totalPhotos <= 1`. Tabular numerals so the digits don't reflow as you click through.
- Adds `photoModal.position` (visible) and `photoModal.positionAriaLabel` (screen reader) translation keys, propagated to all 36 locales via `scripts/sync-translation-keys.js`. Stripped the `[EN]` prefix from the visible numeric format since it's universal; left it on the aria label so the normal auto-translate workflow picks it up.

Closes ticket #620.

## Test plan

- [ ] Open an album with 2+ photos, click any photo, confirm `n / total` appears between the nav arrows.
- [ ] Navigate prev/next — counter updates correctly.
- [ ] Open a single-photo album/page — counter is hidden.
- [ ] Verify it shows on both desktop and mobile (the `.modal-navigation` is in the bottom-right of the modal).
- [ ] Screen reader announces "Photo X of Y" via the aria-label.

### DIFF
--- a/frontend/src/components/ContentModal/ContentModal.tsx
+++ b/frontend/src/components/ContentModal/ContentModal.tsx
@@ -37,8 +37,8 @@ interface ContentModalProps {
 const ContentModal: React.FC<ContentModalProps> = ({
   selectedPhoto,
   album,
-  currentIndex: _currentIndex, // Unused but kept for future UI features
-  totalPhotos: _totalPhotos, // Unused but kept for future UI features
+  currentIndex,
+  totalPhotos,
   onNavigatePrev,
   onNavigateNext,
   onClose,
@@ -691,6 +691,8 @@ const ContentModal: React.FC<ContentModalProps> = ({
               showHint={showNavigationHint}
               onPrevious={handleNavigatePrev}
               onNext={handleNavigateNext}
+              currentIndex={currentIndex}
+              totalPhotos={totalPhotos}
               style={{}}
             />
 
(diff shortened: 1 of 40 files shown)


theodorecharles:

All three acceptance criteria are met. `ContentModal` activates the previously idle `currentIndex` and `totalPhotos` props (removing the underscore aliases and the developer comment), passes them to `ModalNavigation`, which renders `n / total` between the prev/next buttons when `totalPhotos > 1`. The 0-to-1-based index conversion is correct, the counter is styled with tabular numerals to prevent layout shifts while navigating, and all 36 locale files receive the two new translation keys via the existing sync script. No unrelated edits, no new data plumbing, no scope creep.

**Suggestions**
- `frontend/src/components/ContentModal/ModalNavigation.tsx:50` — The `aria-label` is placed on a plain `div` with no ARIA role. Most screen readers ignore `aria-label` on non-interactive, roleless elements, so the intended "Photo X of Y" announcement won't actually be surfaced — assistive tech will just read the visible `X / Y` text directly. Adding `role="img"` (static label) or `role="status"` with `aria-live="polite"` (live update on navigation) would make the described behaviour real. Neither is a blocker for the visual feature, but it's worth a follow-up if screen-reader support was a goal.

PR #237, opened 2026-04-27
feat(admin): multi-select and bulk delete for album photos (ticket #622)

theodorecharles:
## Summary

- Adds a `Select` toggle button to the album content panel header (next to the grid/list view toggle). Activating it switches the grid and list views into multi-select mode: clicking a photo toggles its selection, shift-click on desktop selects a contiguous range, and a sticky bottom action bar appears with Select All, Cancel, and Delete selected.
- Bulk delete shows a single confirmation modal, then issues parallel `DELETE /api/albums/:album/photos/:filename` requests to the existing per-photo endpoint. A toast summarises success/failure counts, and successfully-deleted photos are removed from local state.
- Drag-to-reorder and the per-photo hover overlay are suppressed while in select mode (`useSortable` is disabled, the overlay is hidden via CSS) so the workflows don't conflict. Single-item delete and reorder flows are unchanged when select mode is off.

## Test plan

- [ ] In an album with 3+ photos, click `Select`, click two photos, confirm the bottom bar shows `2 selected` and `Delete selected` is enabled.
- [ ] Click another photo with `Shift` held; confirm the range fills in.
- [ ] Click `Delete selected`, confirm the confirmation modal title shows `Delete N items?`, confirm; toast shows `N photos deleted` and the grid reflects the deletion.
- [ ] Switch to list view, repeat the basic select + delete flow.
- [ ] In select mode, try to drag a photo — drag should not initiate.
- [ ] Exit select mode (press `Done` or `Cancel`); single-item delete via the hover overlay still works.
- [ ] On a touch device, tap-to-select; confirm no overlay appears and selection toggles per tap.


theodorecharles:

The PR satisfies all four acceptance criteria cleanly. The Select toggle lands in both grid and list views, `useSortable` is disabled in both `AlbumGridItem` and `AlbumListItem` while `selectMode` is active, and the single-item delete overlay is suppressed via a `!selectMode` guard — so neither existing workflow is disturbed. Shift-click range selection is implemented correctly with an anchor ref in `handleTogglePhotoSelect`. The bulk delete handler correctly parses `photo.id` (which is `album/filename` throughout the codebase) and issues parallel DELETEs against the existing per-photo endpoint, then calls `exitSelectMode` via the completion callback. One small behavioural difference from the AC: the sticky action bar appears as soon as select mode is entered rather than waiting for the first selection — but the Delete button is correctly disabled at zero selections, and the `selectedCount_zero` string ('Tap photos to select') makes this a better UX. Not a blocker. No automated tests were added, but no CI checks are configured on this branch and there are no pre-existing component tests for the delete flow to pattern-match against, so this is consistent with the project's current testing posture.

PR #239, opened 2026-04-27
i18n: sync albumsManager keys to all 36 frontend locales

theodorecharles:
## Summary
- Batch deploy [workflow_run 25013945155](https://github.com/theodorecharles/Galleria/actions/runs/25013945155) failed translation schema validation: 24 `albumsManager.*` keys added by the album descriptions / bulk-select features in batch #59 existed only in `en.json`.
- Ran `node scripts/sync-translation-keys.js` per `TRANSLATION_WORKFLOW.md` to propagate the keys to all 36 non-EN frontend locales with `[EN]` placeholders. Backend was already in sync.
- `node scripts/validate-translation-schema.js` now reports all 72 frontend + backend files valid — the deploy gate is unblocked.

The `[EN]`-prefixed strings are the documented sync output. `auto-translate.js` needs an OpenAI key from `data/config.json`, which isn't available in this fresh worktree, so the placeholders ship as-is. Ted can run auto-translate locally to fill them; the deploy gate only checks schema, which passes.

Resolves ticket #639.

## Test plan
- [x] `node scripts/validate-translation-schema.js` → all 72 files valid
- [ ] CI on this PR
- [ ] devel deployment (`Update Devel VM`) succeeds where workflow_run 25013945155 failed
- [ ] (optional, post-merge) Ted runs `node scripts/auto-translate.js` locally to replace `[EN]` placeholders with real translations


theodorecharles:

The PR does exactly what the ticket required: 24 `albumsManager.*` keys that existed only in `en.json` (added by the album-description and bulk-select work in batch #59) have been propagated to all 36 non-EN frontend locales using the documented `sync-translation-keys.js` script, and `validate-translation-schema.js` now reports all files valid. The diff is clean — 36 locale files, each getting the same 24 keys appended, no unrelated edits. The `[EN]`-prefixed placeholders are the expected output of the sync script; the workflow doc treats `auto-translate.js` as conditional on having an OpenAI key, and the deploy gate that originally failed checks schema structure, not translation quality. One practical note: CI never ran on this branch (`gh pr checks` returned nothing), so the second acceptance criterion — the workflow passing after merge — can't be confirmed until merge and re-trigger. That's a forward-looking criterion by design, and the fix is straightforward enough that I'm not blocking on it. Ted should run `node scripts/auto-translate.js` locally post-merge to replace the `[EN]` stubs with real translations before the placeholders become user-visible.

**Nits**
- `frontend/src/i18n/locales/bg.json:478` — The `[EN]` placeholders here (and in all 35 sibling files) are the documented intermediate state, but they will surface in the UI for non-English users until `auto-translate.js` is run. Not a blocker — the deploy gate is schema-only — but worth tracking as a follow-up so they don't linger.

PR #240, opened 2026-04-27
fix(admin): use shared ConfirmModalConfig in ModalsCollection (ticket #641)

theodorecharles:
## Summary
- After ticket #639 deployed, `tsc -b` failed in the frontend build with `Property 'title' does not exist on type 'ConfirmModalConfig'` at `ModalsCollection.tsx:994/995`.
- The cause was a stale duplicate: `ModalsCollection.tsx` declared its own local `ConfirmModalConfig` interface that had drifted from the canonical one exported from `../types.ts`. The shared type gained `title?: string` in batch #59 (album-descriptions feature), but the local copy was never updated.
- Fix: delete the local duplicate and import `ConfirmModalConfig` from `../types`. The two shapes are otherwise identical, so this is a no-op at runtime.

## Test plan
- [ ] CI: `tsc -b && vite build` succeeds in the frontend workspace.
- [ ] Manual smoke: open the Admin > Albums view and trigger any flow that opens the generic confirmation modal (delete photo, delete album, etc.) — header text still renders correctly, including any caller that passes an explicit `title`.

### DIFF
--- a/frontend/src/components/AdminPortal/AlbumsManager/components/ModalsCollection.tsx
+++ b/frontend/src/components/AdminPortal/AlbumsManager/components/ModalsCollection.tsx
@@ -9,24 +9,12 @@ import { API_URL } from '../../../../config';
 import { createPortal } from 'react-dom';
 import ShareModal from '../../ShareModal';
 import MoveToFolderModal from './MoveToFolderModal';
-import { Photo, Folder } from '../types';
+import { Photo, Folder, ConfirmModalConfig } from '../types';
 import { cacheBustValue } from '../../../../config';
 import { MagicWandIcon } from '../../../icons';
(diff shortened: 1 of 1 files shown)


theodorecharles:

The root cause is correct and the fix is surgical. Ticket #639 added `title?: string` to the shared `ConfirmModalConfig` in `types.ts` and wired it into the JSX in `ModalsCollection.tsx`, but the file still carried its own local copy of that interface which was never updated. The PR removes the stale duplicate (13 lines gone) and replaces it with a single import from `../types`, where the canonical definition already has `title`. There is no scope creep, no behavioral change, and no test gap — this is a pure type alignment. The acceptance criterion (frontend build completes without TypeScript errors in `ModalsCollection.tsx`) is met.

PR #242, opened 2026-04-28
fix: notify only highest reached milestone per tick (ticket #689)

theodorecharles:
## Summary
- Fixes a bug in `backend/src/services/milestone-tracker.ts` where an album crossing several view-count thresholds in a single tick (e.g. 90 → 5500) would trigger a separate admin push notification — and a separate DB write — for each crossed threshold (100, 1000, 5000).
- Inner loop now iterates `MILESTONES` in descending order and breaks on the first match, so each tick produces at most one notification and one DB update per album, for the highest crossed threshold.
- Cross-tick semantics are unchanged: subsequent ticks still skip already-notified milestones because `updateAlbumMilestone` persists the new value.

## Test plan
- [ ] CI build (`tsc`) passes.
- [ ] Manual: simulate an album whose `lastMilestone = 0` and `views = 5500`; verify exactly one notification ("5,000 views") fires and the DB row is updated to 5000.
- [ ] Manual: with `lastMilestone = 5000` and `views = 12000`, verify a single notification for 10000 fires and DB updates to 10000.
- [ ] Manual: with `lastMilestone = 5000` and `views = 5500` (no crossing), verify no notification.

### DIFF
--- a/backend/src/services/milestone-tracker.ts
+++ b/backend/src/services/milestone-tracker.ts
@@ -31,37 +31,48 @@ export async function checkMilestones(): Promise<void> {
       const currentViews = data.views;
       const lastMilestone = data.lastMilestone;
 
-      // Check if any new milestones were reached
-      for (const milestone of MILESTONES) {
+      // Find the highest milestone the album has crossed since lastMilestone.
+      // Iterating in descending order and breaking on the first match ensures
+      // we send a single notification (and perform a single DB write) per tick
+      // even when an album's view count jumps past several thresholds at once.
+      let highestNewMilestone: number | null = null;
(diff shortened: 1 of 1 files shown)


theodorecharles:

The fix correctly addresses both acceptance criteria. Reversing the iteration order of `MILESTONES` and breaking on the first match yields exactly one notification and one DB write per tick for the highest crossed threshold. The stale `lastMilestone` issue is eliminated by design — there's no longer any local state to refresh. Cross-tick deduplication remains intact because `getAllAlbumViewCounts()` re-reads `lastMilestone` from the DB on every invocation. The change is minimal, touches only the one file identified in the ticket, and introduces no scope creep.

theodorecharles:

PR #243, opened 2026-04-28
fix(setup): handle errors on response/gunzip streams in GeoIP download

theodorecharles:
## Summary

`downloadGeoIPDatabase` in `backend/src/routes/setup.ts` only registered an `error` handler on `fileStream`. If the upstream HTTPS `response` emitted `error` (socket reset mid-download) or `gunzip` emitted `error` (truncated stream / non-gzip body — e.g. db-ip.com returning an HTML error page with status 200), Node would throw an uncaught `'error'` event and the surrounding Promise neither resolved nor rejected — the setup request hung and a partial file was left on disk.

This refactors the two pipe sites (302/301 redirect branch and direct 200 branch) onto `stream/promises.pipeline()`, which was already imported but unused. `pipeline()` propagates errors from every stream in the chain, ensures the destination is closed, and gives one place to unlink the partial file on failure. Also drains (`.resume()`) the redirect-prelude and non-2xx response bodies so their sockets can be released cleanly.

Closes ticket #692.

## Test plan

- [ ] CI passes (typecheck + lint + tests).
- [ ] Manual: run OOBE in a fresh dev env where DB-IP is reachable — confirm `data/GeoLite2-City.mmdb` lands and the existing log lines fire.
- [ ] Manual: simulate failure by pointing `dbipUrl` at an HTML 200 response — confirm `Failed to download GeoIP database stream` logs, no partial file is left, and the setup completes (background promise resolves false instead of hanging).

### DIFF
--- a/backend/src/routes/setup.ts
+++ b/backend/src/routes/setup.ts
@@ -69,58 +69,70 @@ const __filename = fileURLToPath(import.meta.url);
 const __dirname = path.dirname(__filename);
 
 /**
+ * Pipe an HTTPS response through gunzip into a file using stream/promises
+ * pipeline(). pipeline() propagates errors from every stream in the chain
+ * (including socket resets on the response and truncated/non-gzip bodies on
+ * gunzip), ensures the destination is closed, and lets us unlink any partial
+ * file on failure. Returns true on success, false on failure.
+ */
+async function pipeResponseToGunzippedFile(
+  response: import('http').IncomingMessage,
+  dbPath: string
+): Promise<boolean> {
+  const gunzip = createGunzip();
+  const fileStream = createWriteStream(dbPath);
+  try {
+    await pipeline(response, gunzip, fileStream);
+    info('[Setup] GeoIP database downloaded successfully');
+    return true;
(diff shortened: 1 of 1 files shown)


theodorecharles:

The fix correctly addresses all three acceptance criteria. Rather than adding manual `.on('error', ...)` handlers to `response` and `gunzip` (the approach the ticket suggested), the PR routes both pipe sites through `stream/promises.pipeline()`, which was already imported but unused. `pipeline()` propagates errors from every stream in the chain — including socket resets on `response` and decode failures on `gunzip` — and ensures `fileStream` is closed before the `catch` block runs. The `catch` block unlinks the partial file and returns `false`, so the outer `Promise` always resolves cleanly rather than hanging. The bonus `.resume()` calls on redirect-prelude and non-2xx bodies are correct hygiene for socket reuse. The extracted `pipeResponseToGunzippedFile` helper eliminates the duplicated pipe setup between the redirect and direct-200 branches, which is a welcome simplification rather than scope creep. I have no blockers.

theodorecharles:

PR #244, opened 2026-04-28
fix(email): honor per-call language in invitation/reset/test emails (ticket #691)

theodorecharles:
## Summary

The module-scope `t()` helper in `backend/src/email.ts` always called `getGlobalLocale()` and silently dropped the `language` argument that `sendInvitationEmail`, `sendPasswordResetEmail`, and `sendTestEmail` accept. Inviting a French user while the server's global language was English rendered the email in English.

This PR threads the locale through `t(key, vars, locale?)` (default still `getGlobalLocale()` to keep any other callers safe) and adds a small per-function `tr` closure that captures the caller-supplied `language`. Each call site now uses `tr(...)` instead of `t(...)`. The `[Email] ... in language: X` log lines now report the actual per-call language too.

The ticket only named two functions; `sendTestEmail` had the identical bug pattern (also accepts `language: string = 'en'`, also calls `getGlobalLocale()`) and its callers in `routes/config.ts` already pass `siteLanguage`, so it's fixed in the same pass.

No caller-side changes required — `routes/auth-extended.ts` and `routes/config.ts` already pass `siteLanguage` to all three functions.

Closes ticket #691.

## Test plan

- [ ] CI typecheck on the PR
- [ ] Manual: configure SMTP, set global language to `en`, send an invitation to a user whose `language` is `fr` — verify the email body is French
- [ ] Manual: same flow for password reset
- [ ] Manual: trigger a test email from settings; verify it renders in the requested language

### DIFF
--- a/backend/src/email.ts
+++ b/backend/src/email.ts
@@ -27,19 +27,22 @@ export interface EmailConfig {
 
 /**
  * Translation helper function with variable interpolation
- * Replaces {{variable}} placeholders in translated strings
+ * Replaces {{variable}} placeholders in translated strings.
+ *
+ * Pass an explicit `locale` to render in the recipient's language. When omitted,
+ * falls back to the globally configured locale.
  */
-function t(key: string, variables?: Record<string, any>): string {
-  const locale = getGlobalLocale();
-  let translated = translateBackend(key, locale);
-  
+function t(key: string, variables?: Record<string, any>, locale?: string): string {
+  const effectiveLocale = locale ?? getGlobalLocale();
+  let translated = translateBackend(key, effectiveLocale);
+
   // Interpolate variables
   if (variables) {
     Object.entries(variables).forEach(([varName, value]) => {
       translated = translated.replace(new RegExp(`{{${varName}}}`, 'g'), String(value));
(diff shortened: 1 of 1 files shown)


theodorecharles:

The fix threads `language` through correctly on all three counts. The `t()` helper gains an optional `locale` parameter that falls back to `getGlobalLocale()` when omitted — preserving any hypothetical external callers — and each of the three affected functions creates a local `tr` closure that captures its `language` argument and forwards it. Every `t(...)` call site inside `sendInvitationEmail`, `sendPasswordResetEmail`, and `sendTestEmail` is replaced with `tr(...)`. The log lines now report the actual per-call language rather than the global locale, which is a nice bonus. Extending the fix to `sendTestEmail` is sensible scope — same bug pattern, same file, called the same way, and the PR description is transparent about it. No regressions I can see; the fallback path in `t()` keeps the rest of the codebase safe. No automated tests are added, but given this file's style and the manual test plan in the PR body, that's consistent with how the codebase appears to be tested. All three acceptance criteria are satisfied.

theodorecharles:

PR #246, opened 2026-04-28
fix(setup): require one-shot token + admin guard on /api/setup/initialize (ticket #687)

theodorecharles:
## Summary
- Building on #533, this hardens the unauthenticated OOBE flow against takeover even when `config.json` is missing/truncated.
- `refuseIfSetupComplete` now also refuses when an admin user already exists (belt-and-suspenders defense in depth).
- New one-shot setup token: backend generates a 32-byte hex token at startup when no config exists, writes it to `data/setup.token` (mode 0600), and logs it. `/api/setup/initialize` requires the token in the request body and verifies it timing-safely. The token file is deleted on successful init so it cannot be replayed.
- Frontend OOBE wizard polls a new `/api/setup/token-status` endpoint and conditionally renders a setup-token input on the final step. Complete-setup is disabled until the token is filled.
- Existing installs upgrading mid-OOBE (no token file present) still rely on the config + admin guards, so the change is non-breaking.

Closes ticket #687.

## Test plan
- [ ] Fresh install: starting backend with no `data/config.json` writes `data/setup.token` and logs the token.
- [ ] OOBE wizard renders the Setup Token field on step 3, refuses to submit until filled.
- [ ] `/api/setup/initialize` rejects (401) with missing/wrong token; succeeds with the correct one.
- [ ] After successful init, `data/setup.token` is gone and `/api/setup/token-status` returns `{ tokenRequired: false }`.
- [ ] Re-calling `/api/setup/initialize` after init returns 403 (config guard).
- [ ] Manually deleting `data/config.json` post-init still 403s (admin-user guard).

### DIFF
--- a/backend/src/database-users.ts
+++ b/backend/src/database-users.ts
@@ -103,6 +103,33 @@ export function getUserByGoogleId(googleId: string): User | null {
 }
 
 /**
+ * Check whether at least one admin user exists.
+ *
+ * Used by the OOBE setup guard as a belt-and-suspenders check on top of
+ * `getConfigExists()`: if an admin already exists, refuse to re-run the
+ * unauthenticated setup wizard regardless of config.json state. Safe to call
+ * before the users table is created — returns false when the table is missing,
+ * the database is unreachable, or any query error occurs.
+ */
+export function adminUserExists(): boolean {
+  try {
+    const db = getDatabase();
+    // Confirm the users table exists before querying (during OOBE the table
+    // is only created inside /api/setup/initialize).
+    const tableCheck = db
+      .prepare("SELECT name FROM sqlite_master WHERE type='table' AND name='users'")
+      .get();
+    if (!tableCheck) return false;
+    const row = db
+      .prepare("SELECT 1 FROM users WHERE role = 'admin' LIMIT 1")
+      .get();
+    return Boolean(row);
+  } catch {
+    return false;
+  }
+}
+
+/**
  * Create a new user
  */
 export function createUser(data: {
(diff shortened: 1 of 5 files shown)


theodorecharles:

Both acceptance criteria land cleanly. `refuseIfSetupComplete` now double-gates on `getConfigExists()` and the new `adminUserExists()` — so a corrupted or deleted `config.json` cannot re-open the endpoint if an admin row already exists. On top of that, the one-shot setup token gives a second independent layer: a 32-byte hex token is generated at startup (mode 0600, clearly logged), timing-safely verified on every `/initialize` call while the file exists, and deleted before the success response is sent. The frontend polls `/token-status` once on mount and conditionally renders the input and disables the submit button — reasonable for OOBE. The `adminUserExists()` helper safely returns `false` before the `users` table is created, avoiding a bootstrap chicken-and-egg. One observation not worth blocking on: `verifySetupToken` bails early on length mismatch before calling `timingSafeEqual`, which technically leaks token length — but the token is always 64 hex characters so no oracle is reachable in practice. No test suite exists in the repo at all, so the absence of tests here is not a regression introduced by this PR. This goes meaningfully beyond what the ticket required and closes the takeover window correctly.

**Nits**
- `backend/src/utils/setup-token.ts:88` — The early `if (a.length !== b.length) return false` before `crypto.timingSafeEqual` leaks token length via timing. In practice it doesn't matter — the token is always 64 hex chars — but a future refactor that changes token format could make this exploitable. Consider padding both buffers to the same length and always calling `timingSafeEqual`, or just noting the assumption in a comment.

theodorecharles:

PR #247, opened 2026-04-28
feat(admin): inline photo title edit in list view (ticket #695)

theodorecharles:
## Summary
- Double-click a photo title in the admin album list view to edit it in place. `Enter` saves, `Escape` cancels, blur saves.
- Reuses the existing `PUT /api/image-metadata/:album/:filename` endpoint via `usePhotoManagement.updatePhotoMetadata`, threaded as a new `onUpdatePhotoMetadata` prop through `AlbumContentPanel` → `AlbumContentPanelGrid` → `AlbumListItem`. Passes the photo's existing description through unchanged so inline edits never clobber it.
- Pattern mirrors the album-name inline edit in `AlbumContentPanelHeader` (state + focus + select-all on enter, key handlers). The pencil-icon modal still opens the full title-and-description editor for cases that need both fields.

## Test plan
- [ ] Open admin → an album → switch to list view. Double-click a photo title; the row swaps to a focused, all-selected input.
- [ ] Edit and press `Enter` → toast confirms metadata updated, row shows the new title.
- [ ] Edit and press `Escape` → input collapses back to the original title with no save.
- [ ] Edit and click outside (blur) → save fires.
- [ ] Empty title or unchanged title is a no-op (no API call).
- [ ] In multi-select mode, double-click on title does not enter edit mode; row still toggles selection.
- [ ] Drag-and-drop reorder still works on the row (input's pointer events are stopped, but the title text outside edit mode allows drag).
- [ ] An uploading row (status text shown) cannot be edited inline.

### DIFF
--- a/frontend/src/components/AdminPortal/AlbumsManager.css
+++ b/frontend/src/components/AdminPortal/AlbumsManager.css
@@ -1034,6 +1034,37 @@
   text-overflow: ellipsis;
 }
 
+/* Ticket #695 — inline title edit affordance in list view. */
+.title-text.editable {
+  cursor: text;
+  border-radius: 4px;
+  padding: 2px 4px;
+  margin: -2px -4px;
+  transition: background 0.15s ease;
+}
+
+.title-text.editable:hover {
+  background: rgba(255, 255, 255, 0.06);
+}
+
+.list-item-title-input {
+  width: 100%;
+  font-size: 1rem;
+  color: #fff;
+  background: rgba(255, 255, 255, 0.05);
+  border: 1px solid var(--primary-color, #4ade80);
+  border-radius: 4px;
+  padding: 4px 8px;
+  outline: none;
+  font-family: inherit;
+  box-shadow: 0 0 0 3px rgba(74, 222, 128, 0.12);
+}
+
+.list-item-title-input:disabled {
+  opacity: 0.6;
+  cursor: not-allowed;
+}
+
 .status-text {
   font-size: 0.75rem;
   color: #9ca3af;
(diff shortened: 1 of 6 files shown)


theodorecharles:

All four acceptance criteria land cleanly. Double-click on a saved photo title enters edit mode with focus and select-all; Enter saves via the existing `updatePhotoMetadata` PATCH path and collapses back to display; Escape cancels without a network call; the `canInlineEdit` guard correctly blocks edit mode in `selectMode`, during uploads, and for rows without a save handler. The prop is threaded cleanly through `AlbumsManager` → `AlbumContentPanel` → `AlbumContentPanelGrid` → `AlbumListItem`, and `onUpdatePhotoMetadata` is typed optional on `AlbumListItem` so uploading rows continue to work. The implementation follows the `AlbumContentPanelHeader` state pattern faithfully (`isEditingTitle`, `isSavingTitle`, `titleInputRef`, reset effect on external title change). One addition beyond the acceptance criteria: `onBlur` also triggers a save, whereas `AlbumContentPanelHeader` relies on explicit button/key actions only — this creates a slight inconsistency with the "consistent with header" criterion, but it's a net UX improvement for the dense-list context and is called out in the PR test plan. The double-fire guard (`isSavingTitle || !isEditingTitle`) is sound: when Enter fires a save, React re-renders the input as `disabled` before the async resolves, so the unmount-triggered blur sees `isSavingTitle: true` in its closure and exits early. No tests exist for any component in this directory, so there's no test debt being added relative to the codebase baseline. Nothing raises a risk flag.

**Nits**
- `frontend/src/components/AdminPortal/AlbumsManager/components/AlbumListItem.tsx:249` — The `onBlur` saves on click-away, which `AlbumContentPanelHeader` doesn't do. It's arguably better UX for list-scanning workflows, but it means pressing Escape then immediately clicking elsewhere would fire a second cancel/blur cycle (harmlessly, since the guard rejects it). Worth a comment here noting the intentional divergence from the header pattern, so the next reader doesn't 'fix' it back.

PR #248, opened 2026-04-28
feat: cover photo per album (ticket #693)

theodorecharles:
## Summary
- Adds a star "Set as cover" button to the photo overlay in `AlbumGridItem`. Toggles to "Clear cover" when the photo is already the cover. The chosen filename is stored on a new `albums.cover_photo` column.
- Album cards in the admin grid (`SortableAlbumCard`) now render a small cover thumbnail; the social-share OG `preview-grid` (album route and shared-link route) places the cover at index 0 of the 2x2 mosaic. Both fall back to the first photo by sort order when no explicit cover is set.
- Cover is cleared automatically if the underlying photo is deleted. Videos are excluded from cover selection — they already have `VideoThumbnailPicker`.

## What changed

**Backend**
- `database.ts`: idempotent `ALTER TABLE` migration for `albums.cover_photo`; helpers `setAlbumCoverPhoto`, `clearAlbumCoverIfMatches`, `getEffectiveCoverPhoto`. The "effective" helper returns the explicit cover if it still exists in `image_metadata`, else the first photo by sort order.
- `routes/album-management.ts`: new `PATCH /api/albums/:album/cover-photo` (manager-only). Validates the album, validates the filename exists in `image_metadata`, rejects videos, regenerates static JSON, invalidates album cache. Photo DELETE handler now calls `clearAlbumCoverIfMatches`.
- `routes/albums.ts`: `/api/albums` returns `cover_photo` and `effective_cover_photo` on each album.
- `routes/static-json.ts`: per-album JSON payload includes both fields so clients can render cards offline.
- `routes/preview-grid.ts`: `promoteCoverPhoto()` moves the cover to position 0 of the 2x2 mosaic for `/album/:name` and `/shared/:secretKey`. ETag now includes the effective cover so social caches invalidate when it changes.

**Frontend**
- `types.ts`: `Album` gains `cover_photo?: string | null` and `effective_cover_photo?: string | null`.
- `handlers/albumHandlers.ts`: new `handleSetCoverPhoto(albumName, filename | null)` with optimistic local-state update, then PATCH, then `loadAlbums()`.
- Threaded through `AlbumsManager/index.tsx` → `AlbumContentPanel` → `AlbumContentPanelGrid` → `AlbumGridItem`.
- `AlbumGridItem`: star action button (videos hidden) and a star badge that's visible whether or not the overlay is open, so users can see which photo is the cover at a glance.
- `SortableAlbumCard`: renders `${API_URL}/optimized/thumbnail/<album>/<effective_cover_photo>` when present, with `onError` fallback to hide the slot.
- i18n: `setAsCover`, `clearCover`, `coverPhoto`, `coverPhotoSet`, `coverPhotoCleared`, `failedToUpdateCoverPhoto`.

## Test plan
- [ ] Open an album, click the star on a photo — badge appears on that photo, success toast shows.
- [ ] Reopen the album — star badge persists on the same photo.
- [ ] Album card in the admin grid shows that photo as its thumbnail.
- [ ] Click the star on the same photo again — badge disappears, thumbnail falls back to first photo.
- [ ] Delete the cover photo — album silently falls back to first photo (no error).
- [ ] Try clicking the star on a video — button is not rendered.
- [ ] Open the share preview (OG image) for an album with a cover set — the cover is the top-left tile of the 2x2 grid.
- [ ] Existing albums (no cover set) still render correctly with first-photo fallback everywhere.


theodorecharles:

All three acceptance criteria land cleanly. The star button appears in `AlbumGridItem`'s overlay alongside edit and delete, videos are correctly gated out. The `cover_photo` column is added with an idempotent `ALTER TABLE` migration. `SortableAlbumCard` renders the `effective_cover_photo` thumbnail with an `onError` fallback, and `preview-grid.ts` promotes the cover to position 0 in both the `/album/:name` and `/shared/:secretKey` OG mosaics with an ETag that includes the cover filename. The server-side fallback logic in `getEffectiveCoverPhoto` is correct — it verifies the explicit cover still exists in `image_metadata` before trusting it. No scope creep; no unrelated changes.

**Suggestions**
- `frontend/src/components/AdminPortal/AlbumsManager/components/AlbumContentPanelGrid.tsx:289` — The `isCoverPhoto` comparison does `photo.id.split('/').pop()` and compares it directly against `coverPhoto` from the album state. The stored `cover_photo` is saved as a decoded filename (the `AlbumGridItem` handler calls `decodeURIComponent` before passing it to `onSetCoverPhoto`), but `photo.id` might still be URL-encoded if filenames contain spaces or other percent-encoded characters. For most JPEG filenames this won't bite, but it's worth adding a `decodeURIComponent` around the `.pop()` result here to keep the comparison consistent.

**Nits**
- `frontend/src/components/AdminPortal/AlbumsManager/handlers/albumHandlers.ts:729` — The optimistic `effective_cover_photo: filename ?? album.effective_cover_photo` is right when setting a cover, but wrong when clearing — clearing sets `filename` to `null`, so the field keeps the old cover value until `loadAlbums()` resolves. The card thumbnail will briefly still show the cleared photo. Harmless in practice since `loadAlbums()` follows immediately, but the intent of the optimistic update isn't fully realized on the clear path.

PR #251, opened 2026-04-29
feat(admin): "Copy link" button for published albums (ticket #826)

theodorecharles:
## Summary

- Published albums hid both Preview and Share, so the only way to share a live album was to navigate to the public page and copy the address bar. This wires up a single "Copy link" button in the same controls slot that writes `${SITE_URL}/album/<albumName>` to the clipboard and shows a success toast — no modal, no flow.
- Falls back to `document.execCommand('copy')` via a hidden textarea when `navigator.clipboard` isn't available (older browsers / non-secure contexts).
- Adds four new keys under the `albumsManager` namespace (`copyLink`, `copyAlbumLink`, `albumLinkCopied`, `failedToCopyLink`). Ran `scripts/sync-translation-keys.js` so all 36 non-English locales receive `[EN]` placeholders — keeps `validate-translation-schema.js` happy. Auto-translation can be run separately.

## Test plan

- [ ] Open the admin panel, select a **published** album, confirm the new "Copy link" button appears next to the Delete button (where Preview/Share used to live).
- [ ] Click "Copy link" → toast says "Album link copied to clipboard" and pasting yields `https://<your-site>/album/<album-name>`.
- [ ] Album names with spaces / special characters are URL-encoded correctly.
- [ ] Switch to an **unpublished** album: Preview + Share buttons still appear, the new button does not.
- [ ] Resize to mobile width — button still fits in the controls bar (uses existing `photos-btn photos-btn-secondary` styling).
- [ ] `node scripts/validate-translation-schema.js` passes (locally: ✓ 72 files valid).


theodorecharles:

The PR lands exactly what ticket #826 asked for: a single 'Copy link' button in `AlbumContentPanelHeader` that appears only for published albums, copies `${SITE_URL}/album/<albumName>` to the clipboard, and shows a success or error toast. The ternary replacing the old `!isPublished &&` guard correctly preserves Preview + Share for unpublished albums. The URL is built from the same `selectedAlbum` string the rest of the component already uses as the album name, so the construction is sound. Translation keys are propagated to all 36 non-English locales as `[EN]` stubs via the existing sync script, which is the established pattern. Scope is tight — one TSX file and the locale JSONs, nothing unrelated.

**Nits**
- `frontend/src/components/AdminPortal/AlbumsManager/components/AlbumContentPanelHeader.tsx:36` — The `document.execCommand('copy')` call doesn't check its boolean return value, so if the command fails silently (returns `false` rather than throwing) the code still falls through to the success toast. The Clipboard API path is the happy path for every supported browser today, so this rarely matters in practice, but capturing the return and branching to the error toast would make the fallback honest: `const ok = document.execCommand('copy'); if (!ok) throw new Error('execCommand copy failed');`

PR #253, opened 2026-04-29
fix: scope optimization cleanup timer to its own job (ticket #824)

theodorecharles:
## Summary

Fixes a race in `backend/src/routes/image-optimization.ts` where the 5-minute cleanup `setTimeout` for a completed optimization job could null out a *subsequent* job's tracking object.

The reconnect guard at line 188 only blocks while `!isComplete`, so once a job marks itself complete a new POST `/optimize` is allowed to replace the module-level `runningOptimizationJob` (line 217). When the prior job's pending cleanup timer eventually fired it unconditionally set the global to `null`, after which the new job's `child.on('close')` and `child.on('error')` handlers — both guarded by `if (runningOptimizationJob)` — silently skipped broadcasting `complete`/`error`, leaving SSE clients on `/optimize` hung waiting for the terminator and skipping the push notification.

The fix captures the current job in a closure-local `jobRef` and only nulls the global if it still points at the same reference:

```ts
const jobRef = runningOptimizationJob;
setTimeout(() => {
  if (runningOptimizationJob === jobRef) {
    runningOptimizationJob = null;
  }
}, 5 * 60 * 1000);
```

No behavior change in the single-job case.

## Test plan

- [ ] CI build / typecheck passes
- [ ] Manual: start an optimization, let it complete, immediately start a second optimization, then wait 5+ minutes — verify the second job still broadcasts `complete` and that `/status` still reflects the second job correctly while it runs


theodorecharles:

The fix is a textbook closure-capture correction and lands exactly where the ticket said it would. `const jobRef = runningOptimizationJob` is captured inside `child.on('close')` immediately before the `setTimeout`, at the point where the global still unambiguously holds this job's object. The callback then guards the null-assignment with `if (runningOptimizationJob === jobRef)`, which is the precise remedy the ticket prescribed. The `child.on('error')` handler carries no cleanup timer of its own, so no parallel fix is needed there. The diff is one file, ten changed lines, zero unrelated edits. All three acceptance criteria are satisfied: the timer is scoped to its job, SSE clients on a subsequent job will no longer hang, and the push notification path is no longer silently skipped. I have no blockers.

PR #254, opened 2026-04-29
fix(frontend): correct service worker catch handlers to log errors (ticket #823)

theodorecharles:
## Summary

- Fix `TypeError: error is not a function` in `frontend/src/utils/serviceWorker.ts` `.catch()` handlers (lines 76-78 and 96-98).
- The original code called the rejected `error` value as a function — there was no `error` logger imported, so the catch handler always threw and the real SW failure was silently swallowed as an unhandled rejection.
- Imports `error as logError` from `../utils/logger` and renames the catch parameter to `err` to avoid shadowing.

Fixes ticket #823.

## Test plan

- [ ] CI typecheck passes
- [ ] Manually trigger a SW registration failure (e.g. break `/sw.js`) and confirm the logger fires with `[ERROR] Service Worker registration failed: ...` instead of an unhandled rejection.


theodorecharles:

The diff precisely addresses both acceptance criteria from ticket #823. It imports `error as logError` from `../utils/logger`, renames the catch parameter to `err` in both handlers to avoid shadowing, and calls `logError(...)` correctly in both the `registerServiceWorker` and `unregisterServiceWorker` catch blocks. No unrelated edits, no scope creep — one file changed, five lines in, five lines out. CI checks aren't reporting yet, but this is a low-risk, surgical fix and the manual verification step in the test plan is reasonable given the nature of the change.

PR #255, opened 2026-04-29
fix: include toasterSize/isResizing/resizeStart in SSEToaster memo deps (ticket #822)

theodorecharles:
## Summary
- The `useMemo` in `SSEToasterContext.tsx` that builds the provider's context value reads `toasterSize`, `isResizing`, and `resizeStart`, but its deps array omitted all three.
- Consumers received a stale cached value when the user resized or started/stopped a resize — the toaster wouldn't visibly resize, and any UI keyed off `isResizing` was wrong until some other dep (e.g. `toasterPosition`, `isDragging`) changed and forced a recompute.
- Added the three identifiers to the deps array, placed alongside their related state in the same order as the value object.

Fixes ticket #822.

## Test plan
- [ ] CI passes (lint / typecheck / tests).
- [ ] Manual: resize the toaster and confirm size updates immediately, and any `isResizing`-conditional UI reflects the active drag.


theodorecharles:

The diff does exactly what ticket #822 requires: `toasterSize`, `isResizing`, and `resizeStart` are added to the `useMemo` dependency array in `SSEToasterContext.tsx`. I confirmed against the file at HEAD that all three were genuinely absent before this change, and that they appear in the value object the memo builds (lines 223, 231, 233). The additions are placed in the natural order matching the value object — `toasterSize` after `toasterPosition`, then `isResizing` / `resizeStart` after the drag state block — which is clean and readable. Single file, zero deletions, no scope creep. Nothing to flag.

PR #256, opened 2026-04-29
fix(security): validate Origin before reflecting on video routes (ticket #821)

theodorecharles:
## Summary

The HLS handlers (`master.m3u8`, `playlist.m3u8`, segment) and the auth-protected `original.mp4` handler in `backend/src/routes/video.ts` reflected `req.headers.origin` into `Access-Control-Allow-Origin` alongside `Access-Control-Allow-Credentials: true` without consulting the global CORS allowlist. Any third-party site could trigger credentialed cross-origin reads of private/unpublished album video content from an authenticated viewer's session — exfiltrating media that the global CORS middleware would otherwise have blocked.

The fix adds an `isOriginAllowed(origin)` helper to `backend/src/config.ts` that mirrors the rules already used by the global CORS middleware in `server.ts` and the CSRF origin check in `security.ts` (exact match against `getAllowedOrigins()`, then localhost/127.0.0.1, then IPv4 on ports 3000/3001 for Docker/Unraid direct access). Each video handler now gates the CORS header reflection on this helper, and emits `Vary: Origin` when reflecting. The `original.mp4` handler also loses its `Origin || '*'` fallback — wildcard + credentials is invalid per the CORS spec anyway, so that branch was dead code.

Closes ticket #821 (filed by Argus, security audit — MEDIUM severity).

## Test plan

- [ ] CI typecheck/lint passes on PR
- [ ] Same-origin video playback still works for authenticated frontend (Origin matches `FRONTEND_DOMAIN` → reflected)
- [ ] Localhost dev still works (`http://localhost:3000` → reflected)
- [ ] Docker direct access still works (`http://<lan-ip>:3000` → reflected)
- [ ] Hostile origin (`https://evil.example.com`) does NOT receive `Access-Control-Allow-Origin` / `Access-Control-Allow-Credentials` headers — browser blocks the credentialed read


theodorecharles:

The PR correctly fixes all three video handlers named in the ticket and adds a bonus fix for the `original.mp4` handler, which had the same reflected-origin-plus-credentials bug plus a spec-invalid wildcard fallback. The new `isOriginAllowed` helper in `config.ts` faithfully mirrors the three-tier logic already in the global CORS middleware — exact allowlist match via `getAllowedOrigins()`, localhost/127.0.0.1, and IPv4 on ports 3000/3001 — so the validation rules are consistent across the codebase. `Vary: Origin` is added wherever the header is reflected, which is correct. The only minor divergence from the global middleware is that OOBE mode (first-run setup) is not special-cased in `isOriginAllowed`, but that's a non-issue since no authenticated video streaming happens during setup. No unrelated changes, no missing tests beyond what the test plan describes. This is a clean, targeted security fix that satisfies all acceptance criteria.

PR #259, opened 2026-04-30
fix: clean up rejected upload temp files (ticket #887)

theodorecharles:
Fixes ticket #887 by deleting the Multer temp file before returning from post-upload validation failures for invalid album names, rejected filenames, and missing album directories.


theodorecharles:

I read the PR metadata and diff against ticket 887. The change adds a focused temp-file cleanup helper and calls it before the existing invalid album name, invalid filename, and missing album responses, while leaving the successful upload flow unchanged. `gh pr checks` reports no checks configured for this branch, so there is no CI signal to include.

PR #260, opened 2026-04-30
fix: preserve share keys for native hls video (ticket #888)

theodorecharles:
Preserves shared-album secret keys through native Safari HLS playback by keying the master playlist URL and rewriting served playlist URI lines with the share key so variant playlists and segments remain authorized.


theodorecharles:

I read the diff against ticket 888 and the change addresses the Safari/native HLS failure: the frontend now keys the master playlist URL, the HLS.js loader still adds the key for video API requests that do not already have it, and the backend preserves the share key through master and variant playlist URI lines so Safari segment requests remain authorized. I did not find blockers; no CI checks were reported for this branch.

PR #261, opened 2026-04-30
fix: gate preview grids by album access (ticket #886)

theodorecharles:
Enforces the album/media access policy on /api/preview-grid/album/:albumName. Published albums remain public, authenticated sessions can preview private albums, and valid unexpired share links can authorize the album preview via ?key=. Authenticated private preview responses are marked private/no-store.


theodorecharles:

I read the PR diff against ticket #886 and the existing album/video/optimized-media access policy. The changed preview-grid route now validates the album before filesystem reads, denies unpublished albums to unauthenticated users without a valid share key, and still allows published, authenticated, and valid share-link access. I did not find a blocker; GitHub reports no checks for this branch.

PR #263, opened 2026-04-30
fix: redact analytics service token from config (ticket #883)

theodorecharles:
Extends the existing admin-only config redaction path list to include the OpenObserve service token consumed by metrics routes, preserving the placeholder restore behavior for secret rotation.


theodorecharles:

I read the PR against ticket 883. The route is admin-only at the PR head and most sensitive fields are masked, and this patch correctly adds `analytics.openobserve.serviceToken`; however, the OpenObserve basic-auth username remains unredacted even though the ticket explicitly calls out analytics credentials and the backend treats `ANALYTICS_USERNAME` as part of that credential set. I would fix that before merging.

**Blockers**
- `backend/src/routes/config.ts:71` — `analytics.openobserve.username` is still returned in plaintext from `GET /api/config`. The ticket requires analytics credentials to be redacted, and `backend/src/security.ts` treats `ANALYTICS_USERNAME` alongside `ANALYTICS_PASSWORD` as OpenObserve credentials. Add `analytics.openobserve.username` to `SENSITIVE_CONFIG_PATHS` so the admin response and placeholder restore behavior cover the full credential pair.

theodorecharles:

I read the PR diff against the ticket criteria. The PR head already has `GET /api/config` behind `requireAdmin`, returns `maskSensitiveFields(config)`, and restores placeholders on update; this change closes the remaining analytics gap by redacting OpenObserve `username` and `serviceToken` alongside the existing password. I did not find a blocker in the submitted diff. `gh pr checks` reported no checks for the branch.

PR #265, opened 2026-05-02
fix: handle suffix video range requests (ticket #988)

theodorecharles:
Fixes the /original.mp4 range parser so valid suffix-byte requests like Range: bytes=-500 resolve to the tail of the file instead of NaN, and malformed or unsatisfiable ranges return 416 before opening a stream.


theodorecharles:

I read the PR diff against ticket 988. The parser now handles suffix ranges like `bytes=-500`, keeps normal and open-ended ranges working, rejects malformed, inverted, and unsatisfiable ranges with `416`, and does that validation before opening the read stream. I did not find a blocker; GitHub reports no checks for this branch.

PR #266, opened 2026-05-02
fix: require MFA reauth for passwordless accounts (ticket #990)

theodorecharles:
Require sensitive MFA actions to have a fresh proof for passwordless users by accepting a current TOTP code in addition to the existing passkey assertion path, while preserving password checks for credential users.


theodorecharles:

I read the PR diff against ticket 990 and the PR head context. Both destructive MFA routes already route through `verifyReauth`, and this change adds a valid current TOTP proof path while preserving password checks and the existing user-bound passkey assertion path, so OAuth-only users can no longer use only an active session cookie. No CI checks are reported for this branch.

PR #267, opened 2026-05-02
fix: confine video thumbnail upload paths (ticket #982)

theodorecharles:
Sanitizes custom video thumbnail upload route params before deriving output files and rejects resolved thumbnail/modal paths that escape the optimized directory.

### DIFF
--- a/backend/src/routes/album-management.ts
+++ b/backend/src/routes/album-management.ts
@@ -483,6 +483,11 @@ const isVideoFile = (filename: string): boolean => {
   return /\.(mp4|mov|avi|mkv|webm)$/i.test(filename);
 };
 
+const isPathWithinDirectory = (baseDir: string, targetPath: string): boolean => {
+  const relativePath = path.relative(baseDir, targetPath);
+  return relativePath === '' || (!relativePath.startsWith('..') && !path.isAbsolute(relativePath));
+};
+
 /**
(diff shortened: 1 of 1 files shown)


theodorecharles:

I read the diff against ticket 982. The path traversal fix is conceptually in the right area, but the PR introduces calls to `cleanupUploadedTempFile(req.file)` without defining or importing that helper anywhere in `backend/src/routes/album-management.ts` or `backend/src`, so this route will not compile. I would not merge until that is fixed.

**Blockers**
- `backend/src/routes/album-management.ts:1804` — `cleanupUploadedTempFile` is not defined or imported in this file. The new invalid-album, invalid-filename, and invalid-path branches call it, which should fail TypeScript/build before this security fix can ship. Define/import the helper or use the existing local cleanup pattern safely, then rerun checks.

theodorecharles:

I read the PR against ticket #982 and it addresses the requested security fix: decoded album and filename params are validated before use, traversal/path separators are rejected, and the resolved thumbnail/modal outputs are checked against the optimized output root before `sharp().toFile()` writes. The change is narrowly scoped to the vulnerable route and cache/log names now use the sanitized album/filename. There are no reported CI checks and this repo does not appear to have backend route tests to extend, so I do not see a missing-test blocker here.

PR #268, opened 2026-05-02
fix: handle original video stream errors (ticket #989)

theodorecharles:
Adds guarded piping for the /original.mp4 handler so both ranged and full-file streams log read errors, destroy the response, and clean up request-close listeners.


theodorecharles:

I read the diff against ticket #989. The PR is narrow and satisfies the acceptance criteria: both `/original.mp4` branches now attach a stream error handler before piping, errors are logged and destroy the response, and the request close path tears down the file stream while removing the request close listener. No CI checks were reported for this branch.

PR #269, opened 2026-05-02
fix: create MFA attempts table (ticket #987)

theodorecharles:
Creates the missing mfa_attempts table and index for MFA rate limiting, adds an idempotent migration, and changes the recent-attempts query to bind the SQLite lookback modifier instead of interpolating it into SQL.


theodorecharles:

I read the PR diff against the ticket criteria. The change creates `mfa_attempts` in the normal database initialization path and setup path, adds the requested `(user_id, attempted_at)` index, includes an idempotent migration script discoverable by the existing migration runner, and changes `getRecentMFAAttempts` to bind the SQLite lookback modifier instead of interpolating it into SQL. I do not see a blocker; GitHub reports no checks for this branch.

PR #271, opened 2026-05-29
fix: avoid double query separator for exif fetch (ticket #1043)

theodorecharles:
Fixes the EXIF request URL construction so the existing query string is appended directly instead of adding a second question mark. This preserves cache busting and shared-album keys for EXIF requests.

### DIFF
--- a/frontend/src/components/ContentModal/ContentModal.tsx
+++ b/frontend/src/components/ContentModal/ContentModal.tsx
@@ -324,7 +324,7 @@ const ContentModal: React.FC<ContentModalProps> = ({
     try {
       const filename = photo.id.split('/').pop();
       const res = await fetchWithRateLimitCheck(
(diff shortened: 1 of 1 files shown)


theodorecharles:

One-line fix that exactly matches the ticket. `queryString` is built with a leading `?` at lines 77-79, and the old EXIF fetch conditionally prepended a second `?`, producing `/exif??…` — which buried the share-link `key` under an empty/duplicate query key and caused `requireSharedAlbumKey` to 401/404 on shared albums. The change uses `${queryString}` directly, yielding a single `?` separator with `i` and any `key` param intact, consistent with the `${imageQueryString}` direct-append pattern used elsewhere in this file (lines 368, 384, 544). Both acceptance criteria are satisfied. Despite the auth/security-sensitive flags, the diff touches no auth logic — it restores correct key forwarding. No test accompanies it, but for a trivial frontend URL-string fix with no existing test harness around this construction, that's acceptable and not worth blocking.

theodorecharles:

PR #272, opened 2026-05-29
fix: release native-HLS listener and src on VideoPlayer unmount (ticket #1045)

theodorecharles:
## Summary

Fixes a MEDIUM-severity resource leak in `frontend/src/components/ContentModal/VideoPlayer.tsx` (ticket #1045).

On Safari (`canPlayType('application/vnd.apple.mpegurl')`), the effect set `video.src` and attached an anonymous `loadedmetadata` listener, but the cleanup callback only destroyed the hls.js instance — it never removed the listener nor released the media element. Closing the modal mid-load (or switching videos, which re-runs the effect via the `filename` dep) left the element holding the closure and continuing network activity, compounding across opens.

## Changes

- Hoisted the native-HLS listener into a local `onLoadedMetadata` variable so cleanup can `removeEventListener` it.
- Cleanup now releases the media element for both code paths: `video.removeAttribute('src'); video.load();`.

## Test plan

- [ ] Safari: open a video, close the modal mid-load — confirm network activity stops and no stale `loadedmetadata` listener remains.
- [ ] Open multiple videos in sequence — confirm no listener/src accumulation.
- [ ] hls.js (Chrome/Firefox) path still plays and tears down cleanly.

### DIFF
--- a/frontend/src/components/ContentModal/VideoPlayer.tsx
+++ b/frontend/src/components/ContentModal/VideoPlayer.tsx
@@ -48,6 +48,9 @@ const VideoPlayer: React.FC<VideoPlayerProps> = ({
     initializingRef.current = true;
     console.log('[VideoPlayer] Initializing for', filename);
 
+    // Track the native-HLS listener so cleanup can remove it
+    let onLoadedMetadata: (() => void) | null = null;
+
     if (onLoadStartRef.current) onLoadStartRef.current();
 
     const appendSecretKey = (url: string): string => {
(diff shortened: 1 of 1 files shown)


theodorecharles:

Clean, tightly-scoped fix that satisfies all three acceptance criteria. The native-HLS `loadedmetadata` handler is hoisted into a local variable and removed in cleanup against the same stable captured `video` element; cleanup now calls `removeAttribute('src')` and `load()` to release network activity on both unmount and filename change; and the existing hls.js `destroy()` teardown is preserved. Running `load()` after `hls.destroy()` on the detached element is a harmless reset and does not regress the MSE/Chrome/Firefox path. One file, +13/-2, no scope creep, no risk flags. The redundant `onLoadedMetadata = null` and console.log noise are nits not worth blocking on.

theodorecharles:

PR #273, opened 2026-05-29
fix: remove video assets on folder album delete (ticket #1050)

theodorecharles:
Summary:
- Remove per-album video/HLS directories when deleting a folder with deleteAlbums=true.
- Mirror the album deletion cleanup for videoDir so folder deletes do not leave video output behind.

Ticket: #1050


theodorecharles:

I read the PR diff against ticket #1050. The change adds `videoDir` lookup and removes `path.join(videoDir, album.name)` inside the existing `deleteAlbums=true` child-album cleanup loop, preserving the existing photo and optimized cleanup flow. I do not see scope creep or a regression in the touched path. GitHub reports no checks for this branch.

theodorecharles:

PR #274, opened 2026-05-29
fix: trigger SharedAlbum countdown warnings on threshold crossing (ticket #1047)

theodorecharles:
## Summary

The SharedAlbum expiration countdown fired its second-level warnings (`30s`/`10s`/`5s`) using exact-equality on a 1-second `setInterval` (`secondsLeft === 30` etc.). `setInterval` is throttled in background tabs and can drop ticks under load, so `secondsLeft` could jump 31→29 or 11→9 and skip a warning entirely — including the `EXPIRED!!!` toast and the follow-on `setExpired(true)`.

This switches every countdown warning to a **threshold-crossing** check (`prev > N && next <= N`), tracking `prevSecondsLeft` across ticks, so a skipped tick still fires the warning the next time it runs. The once-only `shownWarnings` guard is preserved.

Also fixes a setState-after-unmount bug: the inner `setTimeout(() => setExpired(true), 2000)` was never tracked, so unmounting between EXPIRED detection and the timeout fired `setState` on an unmounted component. It's now stored in a local var and cleared in the effect cleanup alongside the interval.

## Test plan

- [ ] Open a shared album expiring soon in a backgrounded tab; confirm 30s/10s/5s warnings and EXPIRED toast still fire after throttled ticks.
- [ ] Navigate away within 2s of expiry; confirm no setState-after-unmount warning in console.

### DIFF
--- a/frontend/src/components/SharedAlbum.tsx
+++ b/frontend/src/components/SharedAlbum.tsx
@@ -194,6 +194,8 @@ export default function SharedAlbum() {
     }
 
     const shownWarnings = new Set<string>();
+    let prevSecondsLeft = Infinity;
+    let expireTimeout: ReturnType<typeof setTimeout> | undefined;
 
     const checkExpiration = () => {
       const now = Date.now();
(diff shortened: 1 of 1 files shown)


theodorecharles:

The fix does what ticket #1047 asked: all countdown warnings now trigger on threshold crossing (`prevSecondsLeft > N && secondsLeft <= N`) rather than exact equality, so a throttled or dropped `setInterval` tick can't skip the 30s/10s/5s warnings, and the inner `setExpired(true)` timeout is captured in `expireTimeout` and cleared in the effect cleanup — no more setState-after-unmount. The `shownWarnings` once-only guard is intact and ordering the `5s` branch first means a multi-threshold jump fires the most-urgent warning. Clean, single-file, no scope creep. One non-blocking edge case worth a look before merge: initializing `prevSecondsLeft` to `Infinity` makes the immediate on-mount `checkExpiration()` satisfy the crossing condition for any album opened with ≤300s left, firing a spurious/inaccurate warning (e.g. opening with 250s remaining shows '5 minutes left!'). The old exact-equality code didn't do this. Initializing `prevSecondsLeft` from the current `secondsLeft` on the first run, or skipping the warning block on the initial tick, would avoid it.

**Suggestions**
- `frontend/src/components/SharedAlbum.tsx:196` — `prevSecondsLeft = Infinity` plus the immediate `checkExpiration()` on mount means any album opened with ≤300s remaining hits `Infinity > N && secondsLeft <= N` on the first tick and fires a spurious warning (e.g. 250s left → 'five minutes left!'). The else-if chain limits it to one toast, but it's inaccurate and new vs the old exact-equality behavior. Consider seeding `prevSecondsLeft` from the current `secondsLeft` on the first run, or skip the warning block on the initial tick.

theodorecharles:

PR #275, opened 2026-05-29
fix: redirect video page errors after render (ticket #1044)

theodorecharles:
Moves the VideoPage 404 redirect out of the render body and into an effect keyed by the completed error/missing-video state, so React Router navigation is dispatched after commit.


theodorecharles:

I read the diff against ticket 1044. The PR moves the `/404` navigation out of the render path into a `useEffect`, leaves the synchronous `error || !video` render path returning `null`, and keeps the redirect gated until the lookup is no longer loading, which matches the intended behavior. No CI checks are reported for this branch.

theodorecharles:

PR #276, opened 2026-05-29
fix: require video thumbnail for AI titles (ticket #1049)

theodorecharles:
Fixes the AI single-title generator so videos without extracted thumbnails return 422 instead of sending the raw video file to OpenAI Vision. Also derives the image MIME type from the actual file sent, so video thumbnails are sent as JPEGs.


theodorecharles:

I read the diff against the ticket criteria. The PR keeps video inputs on `optimized/thumbnail/<album>/<basename>.jpg`, returns a clear 422 before importing/calling OpenAI or writing `image_metadata` when that thumbnail is missing, and preserves the original fallback path for non-video images. I found no blockers; there were no CI checks reported for the branch and this repo does not appear to have nearby route tests.

theodorecharles:

PR #278, opened 2026-05-29
fix: use data dir for vapid config writes (ticket #1040)

theodorecharles:
Fixes the VAPID key generation route so it persists config.json through the centralized DATA_DIR path instead of deriving a cwd-relative ../data path. This keeps generated keys aligned with Docker volume mounts and DATA_DIR overrides.

### DIFF
--- a/backend/src/routes/push-notifications.ts
+++ b/backend/src/routes/push-notifications.ts
@@ -21,7 +21,7 @@ import {
   sendNotificationToUser,
   type PushSubscription
 } from '../push-notifications.js';
-import { reloadConfig } from '../config.js';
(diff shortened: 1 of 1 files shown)


theodorecharles:

I read the diff against ticket #1040. The PR makes the requested narrow fix: `generate-keys` now imports `DATA_DIR` from `../config.js` and reads/writes `join(DATA_DIR, 'config.json')`, removing the cwd-relative config path from the VAPID key generation route. I did not find scope creep or a correctness issue; GitHub reports no checks for this branch.

theodorecharles:

PR #279, opened 2026-05-29
fix: ContentGrid auto-open matches photo by exact id not endsWith (ticket #1048)

theodorecharles:
## Summary

`?photo=<filename>` auto-open in `ContentGrid` resolved the photo with `allPhotos.find(p => p.id.endsWith(photoParam))`. Since ids are structured `<album>/<filename>`, a filename shared across albums (e.g. `IMG_1234.jpg` in both `vacation/` and `family/`) matched whichever photo came first in array order — not the one the link author meant.

The fix compares the full id on album pages, where the route's `album` equals the photo's source album:

- **Album pages:** `p.id === \`${album}/${photoParam}\`` — exact match, resolves the collision.
- **Homepage:** the feed is sourced from multiple albums and the share link carries only the filename (base `/`), so the album can't be derived from the URL. There we fall back to an exact filename-segment match (`p.id.split('/').pop() === photoParam`), which still removes the `endsWith` partial-suffix false-match risk.

Branching on the existing `isHomepage` flag avoids regressing homepage auto-open, which the naive exact-id fix would have broken.

## Test plan

- [ ] `npx tsc --noEmit` passes (verified locally)
- [ ] Two albums with a same-named photo: `/album/<album>?photo=<file>` opens the correct one
- [ ] Homepage `/?photo=<file>` still auto-opens

### DIFF
--- a/frontend/src/components/ContentGrid.tsx
+++ b/frontend/src/components/ContentGrid.tsx
@@ -155,8 +155,18 @@ const ContentGrid: React.FC<ContentGridProps> = ({ album, onAlbumNotFound, initi
       const urlParams = new URLSearchParams(location.search);
       const photoParam = urlParams.get('photo');
       if (photoParam) {
-        // Find photo by filename
-        const photo = allPhotos.find(p => p.id.endsWith(photoParam));
+        // Find photo by id. Ids are structured `<album>/<filename>`, so we must
+        // compare the album too — otherwise a filename shared across albums opens
+        // whichever photo happens to be first in array order (ticket #1048).
+        // On album pages the route's `album` matches the photo's source album, so an
+        // exact id comparison is correct. The homepage feed is sourced from multiple
+        // albums and the share link carries only the filename, so there we fall back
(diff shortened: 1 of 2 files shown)


theodorecharles:

The change does exactly what the ticket asked. The old `p.id.endsWith(photoParam)` is replaced with an exact-id comparison `p.id === `${album}/${photoParam}`` on album pages, which resolves the cross-album filename collision since ids are structured `<album>/<filename>` (confirmed in `photoHelpers.ts:30`) and the route `album` matches `photo.album` (the share link in `ContentModal.tsx:82-87` builds `/album/${photo.album}?photo=<bare-filename>`). The homepage is correctly branched off via the existing `isHomepage` flag to a filename-segment match, because the homepage share link carries only the filename and its feed spans multiple albums — this avoids regressing homepage auto-open while still eliminating the `endsWith` partial-suffix false match. Both acceptance criteria are satisfied. Scope is tight (one file, +12/-2), and the absence of a test is consistent with the codebase, which has zero frontend test files.

theodorecharles:

PR #280, opened 2026-05-29
fix: require auth on image-metadata GET routes (ticket #1037)

theodorecharles:
## Summary

Closes the unauthenticated data-leak in `backend/src/routes/image-metadata.ts` (security-audit ticket #1037, MEDIUM/auth).

The three GET handlers were mounted with only `csrfProtection` (a no-op for GET requests) and no auth, while the POST/PUT/DELETE handlers on the same router are gated by `requireManager`. This let any anonymous client read titles, descriptions, album names and filenames for **unpublished/private** albums — `/all` returns the entire `image_metadata` table via `db.getAllMetadata()`.

Changes:
- `GET /:album/:filename` → `requireAuth`
- `GET /album/:album` → `requireAuth`
- `GET /all` → `requireManager` (full-table dump, matches the mutating routes)

## Why this is safe for the public gallery

These GETs are not consumed by the public gallery — the only frontend caller is the admin-portal PUT in `usePhotoManagement.ts`. Public image titles are served from the pre-generated static JSON, not these endpoints, so gating them does not affect anonymous browsing.

## Test plan

- [ ] CI typecheck/build passes
- [ ] Authenticated admin/manager can still load metadata in the admin portal
- [ ] Anonymous request to `/api/image-metadata/all` returns 401

### DIFF
--- a/backend/src/routes/image-metadata.ts
+++ b/backend/src/routes/image-metadata.ts
@@ -6,7 +6,7 @@
 import express from 'express';
 import { createRequire } from 'module';
 import { csrfProtection } from '../security.js';
-import { requireManager } from '../auth/middleware.js';
+import { requireAuth, requireManager } from '../auth/middleware.js';
 import { generateStaticJSONFiles } from './static-json.js';
 import { invalidateAlbumCache } from './albums.js';
 import { error, warn, info, debug, verbose } from '../utils/logger.js';
@@ -31,7 +31,7 @@ router.use(csrfProtection);
  * GET /api/image-metadata/:album/:filename
  * Get metadata for a specific image
  */
-router.get('/:album/:filename', async (req, res) => {
+router.get('/:album/:filename', requireAuth, async (req, res) => {
   try {
     const { album, filename } = req.params;
     const db = await getDbFunctions();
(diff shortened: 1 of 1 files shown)


theodorecharles:

Clean, minimal fix that does exactly what the ticket asked for. The three unauthenticated GET handlers now carry middleware: `requireAuth` on `/:album/:filename` and `/album/:album`, and `requireManager` on `/all` to match the full-table dump's mutating siblings. I verified both `requireAuth` and `requireManager` are exported from `backend/src/auth/middleware.js` and the import line was updated correctly. This satisfies both acceptance criteria — anonymous enumeration is blocked, while authenticated/manager access is preserved. The choice of `requireAuth` for the per-resource routes is one of the two remediations the ticket explicitly offered, and the threat model (anonymous enumeration) is closed. No scope creep — the diff is 4 lines changed in a single file. There's no backend test suite to add coverage to, so no missing-test concern. One pre-existing quirk I noticed but won't block on: `/:album/:filename` is registered before `/album/:album` and would shadow it, but that's unchanged by this PR and both routes now share the same auth gate, so no bypass is introduced.

theodorecharles:

PR #281, opened 2026-05-29
fix: harden passkey auth options (ticket #1038)

theodorecharles:
Removes email-based passkey lookup from the unauthenticated passkey auth-options endpoint so responses stay constant-shape and credential IDs are not disclosed. Adds a strict per-IP rate limiter to the endpoint.

### DIFF
--- a/backend/src/routes/auth-extended.ts
+++ b/backend/src/routes/auth-extended.ts
@@ -4,6 +4,7 @@
  */
 
 import { Router, Request, Response } from 'express';
+import rateLimit from 'express-rate-limit';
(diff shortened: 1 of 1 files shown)


theodorecharles:

I read the diff against the ticket criteria. The PR removes the unauthenticated email lookup from `/passkey/auth-options`, stops passing user passkeys into option generation, and adds a per-IP limiter with the same 15-minute/10-request shape used for auth-sensitive routes. I did not find a blocker; no CI checks were reported for the branch.

theodorecharles:

PR #285, opened 2026-05-30
fix: guard AI titles cleanup timer (ticket #1493)

theodorecharles:
Fixes the AI titles completion cleanup race by capturing the completed job reference and only clearing runningJobs.aiTitles if it still points at that same job. This prevents a stale 5-minute timer from orphaning a newer AI titles job's SSE/status/stop tracking.


theodorecharles:

This does exactly what the ticket asked: it captures `const jobRef = runningJobs.aiTitles` and only nulls the global if it still equals `jobRef`, mirroring the #824 fix for image-optimization. Scope is tight — one file, +8/-2, no smuggled refactors — and both acceptance criteria are satisfied: a second job started after the first completes won't be torn down by the first job's stale timer. One residual note worth reading: `jobRef` is captured *after* the two `await translateNotification(...)` calls in the async `close` handler, and `isComplete` is set to `true` before those awaits. That leaves a narrow window where a new POST /generate could replace `runningJobs.aiTitles` during the await, causing `jobRef` to capture the *new* job — the timer would then null the wrong one 5 minutes later. Moving the capture up to right after the `if (runningJobs.aiTitles)` guard (before any await) closes that window fully. Not a blocker; the common case is fixed and this matches the reference pattern. No CI checks are reported on the branch, which is worth a glance before merge.

**Suggestions**
- `backend/src/routes/ai-titles.ts:535` — `jobRef` is captured after the `await translateNotification(...)` calls above, but `isComplete` was set to true before them. During those awaits a new POST /generate can pass the reconnect guard and replace `runningJobs.aiTitles`, so `jobRef` could capture the new job and the timer would null the wrong one. Capture `jobRef` immediately after the `if (runningJobs.aiTitles)` check (before any await) to close that window entirely.

PR #287, opened 2026-05-30
refactor: remove frontend router polling workarounds (ticket #1507)

theodorecharles:
Removes the frontend router polling workarounds from AlbumRoute and NavigationSync, and drops internal navigation fallbacks that manually mutated browser history/location after calling React Router navigate().


theodorecharles:

I read the diff against the ticket criteria. The polling loops in `AlbumRoute` and `NavigationSync` are removed, and the header/footer fallback navigation no longer mutates history or `window.location`. However, the PR still leaves the primes route using `window.location.replace('/primes/')`, which is one of the out-of-router navigation cases called out by the acceptance criteria, so I would not merge this yet.

**Blockers**
- `frontend/src/App.tsx:121` — `PrimesRedirect` still bypasses React Router with `window.location.replace('/primes/')`. The ticket explicitly asks for out-of-router navigations, including primes, to go through React Router `navigate()` instead of `window.location` assignment/replacement. Please convert this redirect flow to router navigation or otherwise address why this route must remain a full page replacement before removing the polling workaround.

PR #288, opened 2026-05-30
perf: debounce and centralize static-JSON regeneration (ticket #1504)

theodorecharles:
## Summary
- Adds `scheduleStaticJSONRegeneration(appRoot)` in `static-json.ts`: a trailing ~750ms debounce with a single in-flight guard plus one coalesced follow-up run, awaited and logged consistently, and invalidating the album cache after each rebuild.
- Replaces all fire-and-forget `generateStaticJSONFiles(appRoot)` call sites — including the per-image and per-video `onComplete` callbacks in `album-management.ts` — with the debounced scheduler. A batch upload of N media now produces one rebuild instead of N, and concurrent unawaited runs no longer race writing the same files under `frontend/dist`.
- Leaves the awaited call sites that report album counts in HTTP responses (publish, homepage, album-description) and the startup regeneration in `server.ts` as direct calls.

## Files
- `backend/src/routes/static-json.ts` (new scheduler + export)
- `backend/src/routes/album-management.ts`
- `backend/src/routes/folder-management.ts`
- `backend/src/routes/image-metadata.ts`

## Test plan
- [ ] Batch-upload several photos to one album; confirm a single regeneration fires ~750ms after the last upload (one `[StaticJSON] Debounced regeneration complete` log) rather than one per photo.
- [ ] Confirm published/homepage album JSON reflects the uploaded media after the debounce window.
- [ ] Confirm admin `/api/static-json/generate` and `/regenerate` endpoints still return album counts synchronously.


theodorecharles:

This does exactly what ticket #1504 asked. `scheduleStaticJSONRegeneration` is a clean trailing-750ms debounce with a proper single in-flight guard (`regenInFlight`) plus a one-slot coalesced follow-up (`regenPending`), and `pendingAppRoot` is refreshed on each call so the trailing run always captures the latest state. I traced the lost-update cases: a mutation arriving during an in-flight run either flips `regenPending` (timer fires while in flight → follow-up run) or re-arms the timer to fire fresh after completion — both paths capture it, so no burst gets dropped. AC1 is met: the per-image and per-video `onComplete` callbacks (the N-rebuilds-per-batch hot path) now schedule instead of regenerating directly. AC2 is met: all 12 fire-and-forget sites across `album-management.ts`, `folder-management.ts`, and `image-metadata.ts` route through the shared scheduler. AC3 is met: errors are logged consistently inside `runRegeneration`, and the completion-dependent sites that report album counts in HTTP responses (publish/homepage/album-description) plus the startup call are correctly left as direct awaited `generateStaticJSONFiles` calls rather than forced through the void-returning scheduler. Scope is tight — 88/-23 across four files, no smuggled refactors. No backend test harness exists in the repo, so the absence of a unit test isn't a convention violation here, though the timing logic would be the one thing worth covering if tests ever land.

**Suggestions**
- `backend/src/routes/static-json.ts:215` — The debounce/coalesce state machine (in-flight guard + trailing follow-up) is the kind of timing logic that's easy to regress silently. No backend test harness exists today so this isn't blocking, but if one is ever added this scheduler is the highest-value thing to cover.

**Nits**
- `backend/src/routes/static-json.ts:184` — `runRegeneration` calls the global `invalidateAlbumCache()` (no arg, clears all albums) after every rebuild, whereas several call sites previously invalidated only their specific album. This is correct — just slightly broader cache churn than strictly needed for the common single-album mutation. Fine to leave; flagging only so it's a conscious choice.

PR #290, opened 2026-05-30
fix: use unique temp filename for uploads to prevent concurrent collisions (ticket #1494)

theodorecharles:
## Summary

Multer's `diskStorage` wrote every upload to `os.tmpdir()/<file.originalname>`, so two concurrent uploads sharing an original name (extremely common for camera files like `IMG_0001.jpg`) opened the **same** temp path with truncation and interleaved bytes — corrupting each other. Even sequentially, a second upload could truncate the temp file the first handler was still reading in the async window before `unlink`.

The fix generates a unique temp filename with `crypto.randomUUID()`, preserving the original extension. `file.originalname` is left untouched, so the downstream handler that derives the final saved name (`sanitizePhotoName(file.originalname)`) is unaffected — only the on-disk temp path is now collision-free.

## Changes
- `backend/src/routes/album-management.ts`: multer `diskStorage.filename` callback now returns `${crypto.randomUUID()}${ext}` instead of `file.originalname`.

## Test plan
- [x] `tsc` clean on the changed line (pre-existing missing-module errors for sharp/openai/dotenv are unrelated)
- [ ] CI on PR
- [ ] Manual: upload two files named `IMG_0001.jpg` concurrently -> both land intact in the album


theodorecharles:
💬 **comment** · recommend: `discuss`

Original reviewer output was unrecoverable: the raw text contained only an upstream API error (400 invalid_request_error about thinking blocks) and no review content could be salvaged.

PR #291, opened 2026-05-30
feat: "Download all" album ZIP for share-link / client galleries

theodorecharles:
## Summary
- Adds `GET /api/albums/:album/download-all` which streams a ZIP of the album's existing **download-size renditions** (the same `optimized/download/<album>/<file>` output the per-photo download serves). No new rendition pipeline — just packages what's already produced.
- Access is gated **identically to the per-photo download**: published album OR authenticated session OR a valid `?key=` share link for that album. Videos are excluded (no download rendition).
- Uses `archiver` with store level (0) — renditions are already JPEG-compressed, so re-compressing is wasted CPU. Streams straight to the response and aborts cleanly if the client disconnects.
- Adds a single **"Download all"** button in `ContentGrid`, which renders on both the public album page and the shared-album view (it passes `?key=` when a share secret is present). Hidden on the homepage and when an album has no downloadable photos.
- New i18n keys `photo.downloadAll` / `photo.preparingDownload` in `en.json`; other locales fall back to English until synced.

## Notes
- New backend dependency: `archiver` (+ `@types/archiver`).
- The button triggers a native streamed download via an anchor click rather than buffering the whole ZIP in memory, so large client galleries don't blow up the browser tab.

## Test plan
- [ ] Published album: "Download all" produces `<album>.zip` containing all photo renditions.
- [ ] Share link (`/shared/:secretKey`): button works with `?key=`; expired/invalid key returns 403/410.
- [ ] Unpublished album with no auth and no key returns 403.
- [ ] Video-only album: no button shown; mixed album excludes videos from the ZIP.


theodorecharles:
💬 **comment** · recommend: `discuss`

Original reviewer output was unrecoverable: the raw text contained an API error message (400 invalid_request_error regarding thinking blocks) rather than any review content. No verdict, findings, or recommendations could be recovered.

PR #293, opened 2026-05-30
refactor: extract shared job runner (ticket #1501)

theodorecharles:
Extracts a shared backend job runner for the long-running optimization flows. Image optimization, video optimization, and the upload optimization stream now share SSE setup, client replay/broadcast bookkeeping, process lifecycle handling, guarded cleanup, and queued child-process execution.


theodorecharles:

I read the diff against the ticket criteria. The PR does centralize SSE client bookkeeping, replay, guarded cleanup, and the upload optimization queue, but it stops short of the main architectural requirement: `JobManager` does not own spawning or child-process lifecycle for the image/video optimization endpoints. Those routes still wire `spawn`, stdout/stderr parsing, close/error completion, and notification timing themselves, so the duplicated lifecycle behavior the ticket asked to extract is still present.

**Blockers**
- `backend/src/services/job-runner.ts:131` — `JobManager` only records an externally spawned child via `setProcess`; it does not own spawning, stdout/stderr handling, close/error handling, or lifecycle callbacks for the long-running image/video jobs. The acceptance criteria call for the shared manager to handle spawn/track of child processes and lifecycle events, with the routes becoming thin consumers. Move the child-process execution/lifecycle into the shared runner API and have `image-optimization.ts` and `video-optimization.ts` register job commands/callbacks instead of each route wiring the process events directly.

theodorecharles:

I read the diff against the ticket criteria. The shared runner extraction mostly covers the duplicated spawn/SSE lifecycle and the album upload queue is indirectly moved through `optimization-stream.ts`, but the queue refactor introduces a state-order regression for per-image optimization jobs, so I would not merge this as-is. CI also reports no checks for this branch.

**Blockers**
- `backend/src/routes/optimization-stream.ts:81` — `QueuedJobRunner.enqueue()` calls `processQueue()` synchronously, so when there is an available worker slot it fires `onStart()` and broadcasts `state: 'optimizing'` before this function later broadcasts the initial `state: 'queued'` at lines 119-125. The previous code pushed the job, broadcast `queued`, then started processing. This can leave clients with a stale queued state after the job is already running, until another progress event arrives. Broadcast the queued state before enqueueing, or change the runner so enqueueing does not start the job until after the caller has emitted the queued event.

PR #294, opened 2026-05-30
refactor: split god-object album-management and auth-extended route files

theodorecharles:
## Summary

Closes #1506. Breaks up the two largest route files into cohesive sub-routers plus a shared module. Handler bodies are moved **verbatim** — no logic changes.

- `album-management.ts` (2078 lines) → `crud`, `photos`, `optimization`, `organization`, `video` sub-routers + `album-management-shared.ts` (helpers, multer config, sanitizers, AI-title functions, path utils).
- `auth-extended.ts` (1853 lines) → `login`, `invites`, `password-reset`, `mfa`, `passkeys`, `users` sub-routers + `auth-extended-shared.ts` (passkey rate limiter, in-memory challenge `Map` singleton, `verifyReauth` + helpers).
- Each original file is now a thin aggregator that mounts its sub-routers on the same base `Router`, so `/api/albums` and `/api/auth-extended` URLs are unchanged. CSRF is still applied once on the album aggregator and therefore covers every sub-router.

## Design notes

- **Flat sibling files in `routes/`, not a subdirectory** — deliberate. Several handlers resolve `projectRoot = path.resolve(__dirname, '../../../')` and use `../` relative imports; moving files into a deeper directory would silently change `import.meta.url`/`__dirname` and break those paths. Keeping the same directory depth preserves them exactly.
- Shared module-level singletons (the passkey `challenges` Map and the rate limiter) live in `auth-extended-shared.ts` so every sub-router references the same instance.

## Test plan

- [x] `npx tsc --noEmit` — no new errors (pre-existing missing `@types` for `speakeasy`/`qrcode`/`bcrypt`/`web-push` remain in untouched files).
- [x] Route-handler count preserved: album-management 15, auth-extended 26.
- [x] Only `server.ts` consumes these files, via default import — unchanged.
- [ ] CI build on the PR.


theodorecharles:

This is a clean, mechanical split. I compared the new modules against the original files at the PR's parent commit and the evidence is solid: the route set (method + path + middleware) is byte-identical for both `album-management` (15 routes) and `auth-extended` (26 routes), and a normalized line-by-line comparison shows zero handler-body lines from the originals are missing in the new modules — every line moved, nothing dropped. The API surface under `/api/albums` and `/api/auth-extended` is unchanged. On the security-sensitive concerns from the risk flags: CSRF is preserved correctly (album applies `csrfProtection` once on the aggregator, which covers all mounted sub-routers; the original `auth-extended` never applied it at the router level, and the new aggregator faithfully omits it too), and the one real refactor hazard — the passkey `challenges` Map and the auth rate limiter — is handled right: both live in `auth-extended-shared.ts` and are pulled in as ES-module singletons, so `passkeys`, `login`, and the reauth path all reference the same instance. No route-ordering shadowing risk since no path patterns overlap ambiguously. I'm satisfied this changes structure without changing behavior.

**Suggestions**
- `backend/src/routes/album-management-crud.ts:1` — The ticket's suggested change asked to route shared side-effects (static-JSON / homepage regeneration) through the centralized helper 'where applicable'. This PR deliberately kept handlers verbatim and didn't rewire those calls. That's a defensible, lower-risk scope decision for a structural split, but it means that part of the suggestion is deferred — worth a quick follow-up ticket rather than blocking this one.

**Nits**
- `backend/src/routes/auth-extended-passkeys.ts:1` — No tests were added or adjusted for the moved auth/album handlers. The acceptance criterion is conditioned on 'existing route tests', and the repo currently has no test files at all, so nothing here regresses — but since this refactor explicitly exists to make these handlers safer to change, a smoke-test harness asserting the 26 auth + 15 album routes still mount and the passkey challenge singleton is shared would be valuable future work.

PR #295, opened 2026-05-30
Replace ad-hoc inline schema migrations with a versioned migration runner

theodorecharles:
## Summary

Closes the architecture-review finding in ticket #1503. `initializeDatabase()` previously evolved the schema with inline `PRAGMA table_info` + conditional `ALTER` blocks wrapped in `try/catch` that swallowed failures into warnings, plus a parallel set of standalone `scripts/migrate-*.js`. There was no `schema_migrations` table, no single source of truth, no ordering guarantee, and a failed migration degraded silently rather than halting startup.

- Adds `backend/src/migrations.ts`: a `schema_migrations` tracking table and an ordered, numbered migration runner. Each migration runs **once**, in order, inside its own transaction, recorded atomically with its effects, and **fails fast** (throws) on error.
- Ports the 7 inline `ALTER` blocks and the DB-schema `migrate-*.js` logic into numbered migrations `001`–`011` (column adds + video backfill, `albums` default `show_on_homepage` recreate, `share_links` `ON UPDATE CASCADE` recreate, `push_subscriptions` table, `.album-milestones.json` import).
- Every migration is defensively idempotent (checks `PRAGMA`/`sqlite_master` before acting), so freshly created and previously upgraded databases — which already have the columns/tables but no `schema_migrations` rows — converge to an identical schema.
- FK enforcement is toggled off around the run (SQLite-recommended for table recreations) and verified with `foreign_key_check` afterwards.
- Removes the 7 ported DB-schema scripts. Config-only `migrate-*.js` (`header-blur`, `header-avatar`, `reset-colors`) are out of scope and still run via `run-all-migrations.js`.

## Test plan

- [x] `tsc --noEmit` clean for the new/changed files (`migrations.ts`, `database.ts`).
- [x] Standalone better-sqlite3 harness covering both **fresh** and **legacy** DBs: fresh applies all 11 and is idempotent on rerun; legacy converges (default flipped to 0, cascade added with data preserved, `.mov` backfilled to video, rename cascades to `share_links`), and rerun is a no-op. `foreign_key_check` reports no violations.
- [ ] CI on this PR runs the authoritative suite.


theodorecharles:

Solid, well-scoped implementation that meets all four acceptance criteria. The runner creates `schema_migrations`, applies numbered migrations in deterministic order each in its own transaction recorded atomically, and fails fast by throwing (aborting startup) instead of degrading to a warning. The 7 inline ALTERs and 4 DB-schema scripts are ported faithfully; the 3 remaining `migrate-*.js` are genuinely config/asset-only and correctly excluded. `run-all-migrations.js` discovers scripts dynamically so the deletions leave no dangling refs, the dropped `warn` import is safe, and FK handling around table recreation (toggle off, recreate, `foreign_key_check`) is correct. Defensive idempotency means a legacy DB's first run is a sequence of no-ops, so fresh and upgraded DBs converge functionally. No blockers — two non-blocking notes below worth a read before merge.

**Suggestions**
- `backend/src/migrations.ts:312` — Convergence is functional, not byte-identical. On a fresh DB, migrations 3 and 5 append `sort_order`/`folder_id` after `created_at`/`updated_at` and migration 8 is skipped, so the column order ends `... created_at, updated_at, sort_order, folder_id`. But when a legacy DB hits this recreate (default was 1), you rebuild as `... description, sort_order, folder_id, created_at, updated_at` — a different physical column order. It's harmless today because every albums query uses named columns (no positional `INSERT INTO albums VALUES` or `SELECT *` index access), but it's a latent footgun. Consider matching the recreate's column order to the fresh-DB order so the two truly converge.
- `backend/src/migrations.ts:426` — The PR description mentions a standalone better-sqlite3 harness exercising fresh + legacy convergence, but it isn't committed. For infra this critical (a failing migration now halts startup), a checked-in regression test covering both paths — fresh applies 1–11 and is a no-op on rerun; legacy converges with data preserved and `foreign_key_check` clean — would protect future migrations from silently breaking convergence.

PR #296, opened 2026-05-30
fix: persist optimization job state (ticket #1502)

theodorecharles:
Persist optimization job lifecycle records in SQLite, reconcile queued/running jobs after restart, and gate image bulk/upload optimization with a shared lock so concurrent writers do not overlap.\n\nVerification: npm run build --prefix backend


theodorecharles:

I read the diff against the reliability criteria. The PR adds durable job rows and a shared in-process lock, but it still misses two restart-facing behaviors the ticket explicitly called out: reconnecting `/optimize` clients are not driven from persisted state, and jobs failed during startup reconciliation do not fire the failure notification path.

**Blockers**
- `backend/src/routes/image-optimization.ts:197` — After a restart, `runningOptimizationJob` is null, so a client reconnecting to `POST /api/image-optimization/optimize` will skip the reconnect path and start a new bulk optimization instead of seeing the reconciled persisted job state. The ticket specifically calls out `/optimize` clients not losing the job after restart; this route needs to consult the persisted latest/running job before creating a new one, and replay/return the reconciled failed or resumable state rather than relying only on module memory.
- `backend/src/database.ts:273` — Startup reconciliation marks queued/running jobs failed directly in the database, but nothing dispatches the corresponding failure notification. Normal in-process completions still notify from route callbacks, but interrupted jobs reconciled here bypass the notification path, so the acceptance criterion that completion and failure notifications still fire from the persisted lifecycle path is not met. Move reconciliation into a lifecycle service that can emit notifications, or have startup collect the interrupted jobs and notify after marking them failed.

theodorecharles:

I read the diff against the ticket criteria. The PR adds durable job records with the requested lifecycle fields, reconciles queued/running jobs on startup by marking them failed and notifying, exposes persisted state back through the optimization streams/status paths, and gates bulk image optimization against queued per-upload optimization with a shared lock. I did not find a merge-blocking correctness issue. There are no GitHub checks reported for the branch, and my temp build attempt could not run because dependencies/tsc are not installed in this workspace.

PR #299, opened 2026-06-25
fix: harden optimized image and setup routes (ticket #1590)

theodorecharles:
Fixes the optimized image cache path traversal by resolving thumbnail/modal cache paths under the optimized root before cache lookup or disk reads, and returns 400 for invalid cache paths. Also locks setup mutation routes after setup is complete and preserves first-run avatar uploads by uploading before initialization.\n\nOperational follow-up is still required: rotate exposed secrets and invalidate sessions because live exposure was confirmed.

### DIFF
--- a/backend/src/routes/setup.ts
+++ b/backend/src/routes/setup.ts
@@ -3,7 +3,7 @@
  * Handles initial setup and configuration for first-time users
  */
 
-import { Router, Request, Response } from "express";
+import { Router, Request, Response, NextFunction } from "express";
 import fs from "fs";
 import path from "path";
 import { fileURLToPath } from "url";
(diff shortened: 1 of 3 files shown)


theodorecharles:

This is a correct, well-scoped fix for the path-traversal vulnerability. `safeOptimizedPath` canonicalizes with `path.resolve`, enforces the `thumbnail/`/`modal/` prefix, and rejects anything resolving outside the optimized root with a 400 before any cache lookup or `fs.readFile` — defeating both raw `../` and single-encoded `%2e%2e` traversal (the ticket's `gallery.db`/`config.json` payloads resolve to `/data/...` and fail the `root + path.sep` check). The `root + path.sep` guard avoids the sibling-prefix bypass, and null bytes/backslashes are rejected. Both `/api/setup/initialize` and `/api/setup/upload-avatar` now get `rejectIfSetupComplete` returning 409, and the frontend correctly reorders the avatar upload before initialize so legitimate first-run avatars still land. All four acceptance criteria are satisfied. No scope creep — the `getSetupState` extraction is a sensible shared-logic refactor. No tests, but the backend has no test suite so that's consistent with the codebase. The PR body correctly flags the required out-of-band secret rotation and session invalidation, which can't be done in code.

**Nits**
- `backend/src/server.ts:585` — Minor inconsistency: `isOptimizedCachePath` accepts a bare `thumbnail`/`modal` (no trailing slash), but `safeOptimizedPath` requires `thumbnail/`/`modal/`, so a request to exactly `/thumbnail` enters the middleware and returns 400 rather than falling through to express.static. Harmless (it's not a real image path), but the two predicates could share the same prefix rule for clarity.

PR #301, opened 2026-06-25
fix: protect unpublished album read endpoints (ticket #1591)

theodorecharles:
Protects public read paths for album metadata, preview grids, EXIF/video metadata, and optimized assets behind a shared published/authenticated/share-link access check. Static albums-data generation now publishes only published albums and removes stale unpublished JSON files. Also tightens share-link secret listing to manager/admin users and adds focused access-policy regression coverage.

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -7,7 +7,7 @@
     "start": "node --no-warnings --loader ts-node/esm src/server.ts",
     "dev": "nodemon --exec 'node --no-warnings --loader ts-node/esm' src/server.ts",
     "build": "tsc",
-    "test": "node --test --loader ts-node/esm src/utils/album-access.test.ts"
+    "test": "TS_NODE_PROJECT=tsconfig.json TS_NODE_TRANSPILE_ONLY=true node --test --loader ts-node/esm src/utils/album-access.test.ts"
   },
   "keywords": [],
   "author": "",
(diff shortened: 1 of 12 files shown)


theodorecharles:

The backend access helper and route wiring mostly address the public leak, but the shared-link EXIF path regresses: after the new backend check, unpublished photo EXIF requires a valid `key`, and the modal fetch still does not send the route secret. I would fix that before merge; route-level regression tests would also reduce risk for this security-sensitive change.

**Blockers**
- `frontend/src/components/ContentModal/ContentModal.tsx:327` — This EXIF request does not include `secretKey`, but shared album URLs keep the secret in `/shared/:secretKey`, not in `window.location.search`. Once the backend denies unpublished EXIF without `?key=...`, valid share-link users will get 403 here. Also, `queryString` already includes a leading `?`, so this currently produces `??...`. Build the EXIF URL from `URLSearchParams`, add `key=${secretKey}` when present, add the cache-bust param once, and append the query string with a single `?`.

**Suggestions**
- `backend/src/utils/album-access.test.ts:54` — These helper tests are useful, but they do not exercise the exposed routes named in the ticket. For a high-risk access-control fix, add route-level regressions for `/api/image-metadata/all`, `/api/image-metadata/album/:album`, `/api/image-metadata/:album/:filename`, `/api/preview-grid/album/:albumName`, EXIF/video metadata, and static JSON cleanup so route ordering and per-endpoint wiring are covered.

theodorecharles:

I read the diff against ticket 1591. The PR applies a shared published/authenticated/share-link album access check across the exposed metadata, preview-grid, album/photo, optimized media, and EXIF/video metadata paths; public static JSON generation is filtered to published albums and cleans stale album JSON; and the share-link listing route is tightened to manager access. I did not find blockers. GitHub reported no checks for this branch.

PR #305, opened 2026-07-16
fix: share FS-first album rename for PUT and PATCH (ticket #2746)

theodorecharles:
## Summary
PATCH `/api/albums/:album/rename` committed album / image_metadata / share_links updates before `fs.renameSync`, and only logged 500 if the photos rename failed — leaving the DB pointing at a name that no longer exists on disk.

The UI already uses PUT (FS first + `renameAlbum()` with FS rollback). This PR collapses both methods onto that safe handler so the live API surface cannot diverge again.

## Changes
- Extract `handleRenameAlbum` from the previous PUT body
- Register `router.put` and `router.patch` on the same handler
- Delete the DB-first PATCH implementation (foreign_keys pragma dance + no rollback)

## Test plan
- [ ] Rename album via admin UI (PUT) — still works
- [ ] `PATCH /api/albums/:album/rename` with valid body — same success shape as PUT
- [ ] Force photos-dir rename failure (e.g. permissions) on PATCH — DB must remain on old name

Ticket #2746

### DIFF
--- a/backend/src/routes/album-management.ts
+++ b/backend/src/routes/album-management.ts
@@ -562,9 +562,11 @@ router.post("/", requireManager, async (req: Request, res: Response): Promise<vo
 });
 
 /**
- * Rename an album
+ * Rename an album (FS first, then DB; roll back FS if DB fails).
+ * Shared by PUT and PATCH so the live API surface cannot diverge into a
+ * DB-first path with no rollback (see ticket #2746).
  */
-router.put("/:album/rename", requireManager, async (req: Request, res: Response): Promise<void> => {
+async function handleRenameAlbum(req: Request, res: Response): Promise<void> {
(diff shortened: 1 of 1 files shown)


theodorecharles:

approve — both AC bullets are met: PATCH `/:album/rename` no longer uses the DB-first path; it shares `handleRenameAlbum` with PUT (FS rename first, then `renameAlbum()`, FS rollback if DB fails), so a photos-dir rename failure leaves albums/image_metadata/share_links on the old names. Dead PATCH body (foreign_keys pragma dance, no rollback) is deleted; only `backend/src/routes/album-management.ts` changed; no mid-thread scope adds to satisfy.

theodorecharles:
Merged in batch #322.

PR #306, opened 2026-07-16
fix: reject getVideoMetadata on spawn failure and empty streams (ticket #2742)

theodorecharles:
## Summary
`getVideoMetadata` only handled `close`/`data` on the ffprobe child. If ffprobe was missing (ENOENT) the Promise never settled, so video upload/processing hung. On success with empty `streams[]`, property access on `undefined` threw.

## Changes
- `ffprobe.on('error', …)` rejects the Promise (matches other spawn sites in this module)
- Guard `!data.streams?.[0]` before reading dimensions/duration/rotation

## Test plan
- [ ] Process/upload path with ffprobe absent rejects promptly (no hang)
- [ ] File with no video stream rejects with a clear error
- [ ] Normal video still returns width/height/duration/rotation

### DIFF
--- a/backend/src/utils/video-processor.ts
+++ b/backend/src/utils/video-processor.ts
@@ -75,6 +75,11 @@ export async function getVideoMetadata(videoPath: string): Promise<VideoMetadata
       stderr += data.toString();
     });
 
+    // Reject on spawn failure (e.g. ffprobe missing / ENOENT) so callers do not hang
+    ffprobe.on('error', (err) => {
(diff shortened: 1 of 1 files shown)


theodorecharles:

approve — both AC bullets are in the diff: `ffprobe.on('error', …)` rejects on spawn failure so callers no longer hang, and empty `streams` is guarded with a clear reject before dimension access. Single-file change in `backend/src/utils/video-processor.ts`; no mid-thread scope adds missing.

theodorecharles:
Merged in batch #322.

PR #307, opened 2026-07-16
fix: prevent optimization timeout double-firing onError/processQueue (ticket #2745)

theodorecharles:
## Summary
Optimization jobs that hit the 5-minute timeout were calling `onError` and `processQueue` twice: once in the timeout handler and again when the killed child emitted `close`. That double-broadcast SSE errors and could start two queued jobs when capacity was freed twice, overshooting `MAX_CONCURRENT_JOBS`.

## Change
- Gate timeout / close / error through a single `finishJob` helper with a `finished` flag in `backend/src/routes/optimization-stream.ts`.
- First terminal path wins; subsequent close after SIGTERM is a no-op.

## Test plan
- [ ] Timeout path: hung optimize job past 5m → one error broadcast, one queue drain
- [ ] Normal success/failure exit still completes once
- [ ] Spawn error path still reports once without double processQueue

Ticket #2745

### DIFF
--- a/backend/src/routes/optimization-stream.ts
+++ b/backend/src/routes/optimization-stream.ts
@@ -109,13 +109,27 @@ async function processQueue() {
 
     activeJobs.add(childProcess);
 
-    // Add timeout to prevent hung processes (5 minutes)
-    const timeout = setTimeout(() => {
-      error(`[OptimizationStream] Job ${job.jobId} timed out after 5 minutes, killing process`);
-      childProcess.kill('SIGTERM');
+    // Only one terminal path may call onError/onComplete + processQueue.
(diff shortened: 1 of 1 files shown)


theodorecharles:

approve — both AC bullets are met in `backend/src/routes/optimization-stream.ts`: timeout, close, and error share a single `finishJob` helper with a `finished` flag so the first terminal path wins; after timeout kill, close is a no-op (no second `activeJobs.delete`, `onError`, or `processQueue`). No mid-thread scope adds beyond that fix; change is scoped to the double-fire bug only.

theodorecharles:
Merged in batch #322.

PR #308, opened 2026-07-16
fix: require admin and CSRF for system restart (ticket #2740)

theodorecharles:
## Summary
- Gates `POST /api/system/restart` with `requireAdmin` instead of `requireAuth` so viewers/managers cannot take the stack down.
- Adds `csrfProtection` on the same route, consistent with other mutating admin routes.

## Context
Felix finding: any logged-in user could hit restart and trigger `pm2 restart all` or `process.exit(0)`. Middleware docs treat system settings as admin-only; managers are content-only, viewers read-only.

## Test plan
- [ ] As admin: restart from config UI still works (`RestartModal` → 200).
- [ ] As manager/viewer: `POST /api/system/restart` with session cookies returns 403.
- [ ] Unauthenticated: 401.
- [ ] Cross-origin POST without allowed Origin: 403 from CSRF.

### DIFF
--- a/backend/src/routes/system.ts
+++ b/backend/src/routes/system.ts
@@ -8,7 +8,8 @@ import { spawn } from 'child_process';
 import path from 'path';
 import fs from 'fs';
 import { fileURLToPath } from 'url';
-import { requireAuth } from '../auth/middleware.js';
+import { requireAdmin } from '../auth/middleware.js';
+import { csrfProtection } from '../security.js';
 import { error, warn, info, debug, verbose } from '../utils/logger.js';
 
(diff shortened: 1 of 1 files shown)


theodorecharles:

approve — all three AC bullets land in `backend/src/routes/system.ts`: `POST /restart` is now `csrfProtection, requireAdmin` (not `requireAuth`), so viewers/managers get 403 via role check and unauthenticated still 401; handler body is unchanged so admins keep the same restart path; CSRF matches other mutating admin routes (e.g. notification-preferences). No mid-thread scope adds missing; no scope creep.

theodorecharles:
Merged in batch #322.

PR #309, opened 2026-07-16
fix: unlink multer temp file on rejected upload paths (ticket #2744)

theodorecharles:
## Summary
After multer `diskStorage` accepts an upload into `os.tmpdir()`, the `POST /:album/upload` handler could reject invalid filenames, missing albums, or invalid album names with 400/404 and return without unlinking `req.file.path`. The outer catch had the same leak. Large temp files could accumulate on every rejected attempt.

## Changes
- Add `unlinkTempUpload` helper (best-effort `fs.unlinkSync`, ignore cleanup errors)
- Call it on early returns: invalid album name, invalid filename, album not found
- Call it in the outer catch of the upload handler
- Success and processing-error paths already unlinked; left as-is

## Ticket
#2744

## Test plan
- [ ] Upload with invalid filename → 400 and no leftover file under system temp
- [ ] Upload to non-existent album → 404 and temp removed
- [ ] Valid upload still succeeds (temp cleaned after move/process as before)

### DIFF
--- a/backend/src/routes/album-management.ts
+++ b/backend/src/routes/album-management.ts
@@ -432,6 +432,19 @@ const sanitizeName = (name: string): string | null => {
 };
 
 /**
+ * Best-effort cleanup of a multer diskStorage temp file.
+ * Safe to call when the path is missing or already unlinked.
+ */
+const unlinkTempUpload = (filePath: string | undefined | null): void => {
+  if (!filePath) return;
+  try {
+    fs.unlinkSync(filePath);
+  } catch {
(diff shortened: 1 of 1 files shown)


theodorecharles:

approve — all four AC bullets are covered in `backend/src/routes/album-management.ts`: invalid filename and missing album early returns call `unlinkTempUpload` before 400/404; outer catch unlinks `req.file?.path`; success and in-handler processing-error paths still unlink as before. Invalid album name is cleaned too. No mid-thread scope adds; no unrelated churn.

theodorecharles:
Merged in batch #322.

PR #311, opened 2026-07-16
fix: reject deactivated users in requireAuth (ticket #2738)

theodorecharles:
## Summary
- `requireAuth` now checks `user.is_active` after the existing DB existence check (Passport + credential sessions).
- Inactive users get the same treatment as deleted ones: session destroyed, `401` with `Account is disabled` (matches login).

## Context
Deactivating a user (`is_active=0`) while they already have a session left protected routes fully open. Login already blocked inactive accounts; middleware did not.

## Test plan
- [ ] Log in as a user, then set `is_active=0` for that user in the DB
- [ ] Call any `requireAuth` route with the existing session cookie → expect `401` and session cleared
- [ ] Confirm active users still pass `requireAuth` normally
- [ ] Confirm login still rejects inactive users with the same error string

Ticket #2738

### DIFF
--- a/backend/src/auth/middleware.ts
+++ b/backend/src/auth/middleware.ts
@@ -30,7 +30,7 @@ export function requireAuth(req: Request, res: Response, next: NextFunction) {
   if (req.isAuthenticated && req.isAuthenticated()) {
     const sessionUser = req.user as any;
     
-    // Verify user still exists in database
+    // Verify user still exists and is active in database
     if (sessionUser?.email) {
       const dbUser = getUserByEmail(sessionUser.email);
       if (!dbUser) {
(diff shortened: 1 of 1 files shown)


theodorecharles:

approve — both AC bullets are in the diff: `requireAuth` destroys the session and returns 401 `Account is disabled` when `!dbUser.is_active` on the Passport and credential paths (same string as login in `auth-extended.ts` / `config.ts`), and active users still reach `next()` with no other behavior change. Scoped to `backend/src/auth/middleware.ts` only; no mid-thread scope adds to fold in.

theodorecharles:
Merged in batch #322.

PR #312, opened 2026-07-16
fix: log SW registration failures without shadowing error (ticket #2743)

theodorecharles:
## Summary
Service worker catch handlers bound the rejection as `error` then called `error(...)` as if it were the logger. Only `info` was imported, so a real SW registration failure became a TypeError in the catch path.

## Changes
- Import `error as logError` from the frontend logger (same pattern as AuthContext / SetupWizard).
- Use `logError` in register and unregister `.catch` handlers so the original failure is logged.

## Test plan
- [ ] Confirm catch paths no longer call the rejection value as a function.
- [ ] Optional: force SW register failure on an HTTPS host and verify the log message appears without a secondary TypeError.

### DIFF
--- a/frontend/src/utils/serviceWorker.ts
+++ b/frontend/src/utils/serviceWorker.ts
@@ -3,7 +3,7 @@
  * Registers and manages the service worker for caching
  */
 
-import { info } from '../utils/logger';
+import { info, error as logError } from '../utils/logger';
 
 export function registerServiceWorker() {
   // Disable service worker on localhost (development)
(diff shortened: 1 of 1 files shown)


theodorecharles:

approve — AC met: `frontend/src/utils/serviceWorker.ts` imports `error as logError` and both register/unregister `.catch` handlers call `logError(...)` with the rejection param, so SW failures log instead of throwing TypeError. No mid-thread scope adds; single-file fix, no creep.

theodorecharles:
Merged in batch #322.

PR #313, opened 2026-07-16
fix: append share key for Safari native HLS (ticket #2741)

theodorecharles:
## Summary
Safari/iOS native HLS set `video.src` to the master playlist without `?key=`, so unpublished album share viewers got 403 from `hasAlbumAccess`. HLS.js already appended the key via `CustomLoader`.

- Frontend: append `?key=` on the native HLS branch when `secretKey` is present.
- Backend: when serving master/media m3u8 with a share key, rewrite relative URI lines to carry `?key=` (Safari does not propagate query params to segments). Key-bearing responses use `Cache-Control: private, no-store`.

## Test plan
- [ ] Open an unpublished album share link on Safari/iOS and play a video — master, variant playlist, and segments should authorize.
- [ ] Same share link on Chrome/desktop (HLS.js path) still plays.
- [ ] Authenticated/admin video playback unchanged (no share key).
- [ ] Published album video playback unchanged.

### DIFF
--- a/backend/src/routes/video.ts
+++ b/backend/src/routes/video.ts
@@ -26,21 +26,76 @@ const sanitizePath = (name: string): string | null => {
 };
 
 /**
+ * Extract share key from request query (validated hex format).
+ */
+const getShareKeyFromQuery = (req: Request): string | null => {
+  // Note: req.query.key can be a string OR an array if multiple keys are provided
+  const shareKeyParam = req.query.key;
+  const shareKey = Array.isArray(shareKeyParam) ? shareKeyParam[0] : shareKeyParam;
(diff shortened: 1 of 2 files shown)


theodorecharles:

approve — AC is met: native Safari path in `VideoPlayer.tsx` appends `?key=` when `secretKey` is set, and `backend/src/routes/video.ts` rewrites master/media m3u8 URI lines under a share key so variant playlists and segments keep authorization (Safari does not propagate query params). Mid-thread scope add is in the same diff; key-bearing playlists use `Cache-Control: private, no-store`. No blockers.

theodorecharles:
Merged in batch #322.

PR #314, opened 2026-07-16
fix: reload role from DB in requireManager/requireAdmin (ticket #2737)

theodorecharles:
## Summary
`requireManager` / `requireAdmin` trusted `session.user.role` frozen at login. Demoting a manager to viewer left their cookie able to hit content/admin APIs until re-login.

## Change
- `getUserFromRequest` always loads `role` and `is_active` from the DB (credential `userId`, Passport email/id, defensive session.user fallback).
- Privileged middlewares reject inactive accounts with 401.

## Test plan
- [ ] Login as manager, keep session, admin demotes to viewer via PATCH role — next manager-only API returns 403 without re-login
- [ ] Admin role still passes `requireAdmin` after a normal session
- [ ] Google OAuth admin/manager still authorized after role is set in DB
- [ ] Deactivated user with live session fails privileged checks

### DIFF
--- a/backend/src/auth/middleware.ts
+++ b/backend/src/auth/middleware.ts
@@ -76,40 +76,72 @@ export function requireAuth(req: Request, res: Response, next: NextFunction) {
 export const isAuthenticated = requireAuth;
 
 /**
- * Helper function to get user from request with role lookup from database
+ * Build privileged-check user from a live DB row.
+ * Role and is_active must never come from session snapshots (frozen at login).
+ */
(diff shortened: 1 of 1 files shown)


theodorecharles:

approve — both AC bullets are met: `getUserFromRequest` always reloads `role` and `is_active` from the DB (credential `userId`, Passport, and defensive `session.user` paths) via `userFromDb`, so demotion to viewer denies `requireManager`/`requireAdmin` without re-login; inactive accounts get 401 on those checks. No mid-thread scope adds; change is scoped to `backend/src/auth/middleware.ts` with no unrelated edits.

theodorecharles:
Merged in batch #322.

